PR #5, opened 2017-12-25
demo

daotianya:
coding


xtaci:
it seems that you missed the implementation of MerkleTree class , which should include GetMerkelProof function

PR #7, opened 2018-01-02
Add asynchronous schedule strategy

shiwk:


xtaci:
thanks for the contribution, can you provide more comments to the functions?
a detailed explanation on the steps of the algorithm is prefered.

shiwk:
we use undirected graph to process Transaction Dependence. DFS is applied to traverse graph to  execute and remove the node with most neighbors Synchronously. Repeat above process until the graph is split apart. Each subgraph is processed Asynchronously.  And we are now working on algorithm optimization.

xtaci:
@shiwk can you put some words directly to the source code, like /*  we use....  */

shiwk:
@xtaci yes, of course

xtaci:
that would be convenient for developers to understand

PR #21, opened 2018-01-08
Implement the merkle proof list.

EanCuznaivy:
Given a merkle tree and a leaf to proof, return a list of proof nodes.
Also add a test case.


xtaci:
cool, is it works for proving the hash list now ?

EanCuznaivy:
@xtaci Yeah, the method VerifyProofList() in MerkleTree.cs can verify a given proof list by a series of hash computes.
Also add a test case VerifyProofListTest().

xtaci:
great job! thanks for the pr!

PR #24, opened 2018-01-11
Working on implementation of minimal blockchain

EanCuznaivy:
Still have a lot to learn.


AElf.Kernel/ExtensionMethods.cs
```diff
@@ -6,9 +6,15 @@ namespace AElf.Kernel
 {
     public static class ExtensionMethods
     {
-        public static byte[] ComputeHash(this byte[] buffer)
+        public static byte[] GetSHA256Hash(this ITransaction tx)
```
loning: ITransaction has a GetHash method, you should use it

xtaci:
GetXXX is a readonly method

PR #26, opened 2018-01-13
IWorker Interface and UnitTest for scheduler

shiwk:
add Interface IWorker with simple implement and working on UnitTest work


xtaci:
A worker is an abstraction of the entity who process the transactions, I might be a async Task or a remote actor in Akka.net

xtaci:
I think a TPL based real worker should be embed into the kernel.

PR #28, opened 2018-01-16
Re-org the codes about blocks

EanCuznaivy:
Tuned some properties

### DIFF
--- a/AElf.Kernel/Block.cs
+++ b/AElf.Kernel/Block.cs
@@ -11,6 +11,7 @@ namespace AElf.Kernel
     [Serializable]
     public class Block : IBlock
     {
+        #region Seems useless for now, but maybe useful in the future
         public int MagicNumber => 0xAE1F;
 
         /// <summary>
@@ -19,20 +20,34 @@ namespace AElf.Kernel
         /// BlockHeader: 84B
         /// </summary>
         public int BlockSize => 92;
+        #endregion
 
+        #region Private fields
         private BlockHeader BlockHeader { get; set; }
-
         private BlockBody BlockBody { get; set; } = new BlockBody();
+        #endregion
 
+        /// <summary>
+        /// When we want to generate a new block,
+        /// we must now know the hash value of previous block.
+        /// </summary>
+        /// <param name="preBlockHash"></param>
         public Block(Hash<IBlock> preBlockHash)
         {
             BlockHeader = new BlockHeader(preBlockHash);
         }
 
+        /// <summary>
+        /// Add a transaction to block body.
+        /// </summary>
+        /// <param name="tx"></param>
+        /// <returns></returns>
         public bool AddTransaction(ITransaction tx)
         {
             if (BlockBody.AddTransaction(tx))
             {
+                //If successfully add a transaction to the block body,
+                //add the hash value of transaction to block header.
                 BlockHeader.AddTransaction(tx.GetHash());
                 return true;
             }
--- a/AElf.Kernel/BlockHeader.cs
+++ b/AElf.Kernel/BlockHeader.cs
@@ -1,32 +1,26 @@
-﻿using Newtonsoft.Json;
-using System;
-using System.IO;
-using System.Runtime.Serialization;
-using System.Runtime.Serialization.Formatters.Binary;
-using System.Security.Cryptography;
-using System.Text;
+﻿using System;
 
 namespace AElf.Kernel
 {
     [Serializable]
     public class BlockHeader : IBlockHeader
     {
-        public int Version => 0;
-
+        /// <summary>
+        /// The hash value of pervious block.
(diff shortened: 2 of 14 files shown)


AElf.Kernel/Block.cs
```diff
@@ -19,20 +20,34 @@ public class Block : IBlock
...
         {
             BlockHeader = new BlockHeader(preBlockHash);
         }
 
+        /// <summary>
+        /// Add a transaction to block body.
+        /// </summary>
+        /// <param name="tx"></param>
+        /// <returns></returns>
         public bool AddTransaction(ITransaction tx)
```
loning: we need to add tx id to block header merkle tree

AElf.Kernel/BlockHeader.cs
```diff
@@ -1,32 +1,26 @@
...
-                return GetTransactionMerkleTreeRoot();
-            }
-        }
-
-        private MerkleTree<ITransaction> MerkleTree { get; set; } = new MerkleTree<ITransaction>();
-
+        /// <summary>
+        /// Time stamp.
+        /// </summary>
         public long TimeStamp => DateTime.UtcNow.Second;
```
loning: DateTime.UtcNow.Second is not right

AElf.Kernel/BlockHeader.cs
```diff
@@ -42,16 +36,32 @@ public BlockHeader(IHash<IBlock> preBlockHash)
         /// </summary>
         public int Nonce { get; set; }
 
+        #region Private fields
+        private MerkleTree<ITransaction> TransactionTrie = new MerkleTree<ITransaction>();
```
loning: private filed should be _transactionTrie, or _merkleTreeOfTransactions

AElf.Kernel/BlockHeader.cs
```diff
@@ -62,8 +72,12 @@ public IHash<IBlockHeader> GetHash()
             return new Hash<IBlockHeader>(this.GetSHA256Hash());
         }
 
-        public void SetNonce()
+        /// <summary>
+        /// Adjust the value of Nonce while mining.
+        /// </summary>
+        public void AdjustNonceWhileMining()
```
loning: do not add any mining method in the implement of Block

AElf.Kernel/BlockHeader.cs
```diff
@@ -62,8 +72,12 @@ public IHash<IBlockHeader> GetHash()
             return new Hash<IBlockHeader>(this.GetSHA256Hash());
```
loning: Should just this.GetHash(), not GetSHA256Hash

AElf.Kernel/ITransactionReceiver.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AElf.Kernel
+{
+    public interface ITransactionReceiver
```
loning: we do not need ITransactionReceiver in Kernel

AElf.Kernel/ITransactionSender.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace AElf.Kernel
+{
+    public interface ITransactionSender
```
loning: we do not need ITransactionSender

AElf.Kernel/LiteDBDataProvider.cs
```diff
@@ -23,29 +23,29 @@ public byte[] Serialize()
...
             task.Start();
             return await task;
         }
 
         async Task IAccountDataProvider.SetAsync(IHash key, ISerializable obj)
         {
-            var c = this.db.GetCollection<Record>("data");
-            Task<bool> task = new Task<bool>(() => c.Upsert(new Record(){Key = key, Value = obj.Serialize()}));
+            var c = db.GetCollection<Record>("data");
+            Task<bool> task = new Task<bool>(() => c.Upsert(new Record() { Key = key, Value = obj.Serialize() }));
```
loning: use Task.Factory to start new task

AElf.Kernel/Miner.cs
```diff
@@ -7,8 +7,8 @@ public byte[] Mine(IBlockHeader blockheader)
             int bits = (blockheader as BlockHeader).Bits;
             while (true)
             {
-                //Change the nonce
-                (blockheader as BlockHeader).SetNonce();
+                //Change the Nonce
+                (blockheader as BlockHeader).AdjustNonceWhileMining();
```
loning: AdjustNonceWhileMining should be implement in some class implemented IMiner

AElf.Kernel/BlockHeader.cs
samfrb: This is purely coding convention, but I agree with it so I will put it in the style guide.

PR #43, opened 2018-02-06
Type change : the From and To fields from Transaction are now byte ar…

samfrb:
I was writing serialization code for transaction and noticed that **From** and **To** are of type **IAccount**, but it needs to be a byte array since the only info the sender of a transaction has, is the address (a byte array) that represents the account.
This also means that the interface from the World State as to be slightly modified to be able to get the account data provider from a byte array.


AElf.Kernel/ITransaction.cs
```diff
@@ -19,28 +19,28 @@ public interface ITransaction
...
         /// </summary>
         object[] Params { get; set; }
-        
+
         /// <summary>
         /// The caller
         /// </summary>
-        IAccount From { get; set; }
-        
+        byte[] From { get; set; }
```
loning: IAccount is ok, because IAccount is seralizable

AElf.Kernel/Account.cs
```diff
@@ -7,15 +7,21 @@ namespace AElf.Kernel
     /// </summary>
     public class Account : IAccount
     {
-        private Address m_address;
+        private readonly byte[] _address;
```
loning: Should be IHash<IAccount>

loning:
thx all

PR #59, opened 2018-03-08
some modifications about SmartContractZero and Chain initialization

shiwk:


shiwk:
"var smartContract = await runner.RunAsync(reg)"  the way I implement is directly create SmartContract using SmartContractRegistration, which needs the contract class written by user should implement ISmartContract just like the Contract Demo in AElf.Kernel.Test.

shiwk:
another point to be confirmed is that in my design for now, when a new contract deployed, a new account address will be generated which is associated with this contract. BTW, SmartContractRegistration is created in kernel, and hash of that is calculated with contract name and account who deploy it. And the hash value is also the address of new account.  The hash calculating method can be changed later.

PR #60, opened 2018-03-11
Minor spelling mistakes fix across two README files

dennisfurrer:
The repository's main README file had "demostration" which has been corrected to "demonstration".
The AElf.Kernel README had "Tranaction" which has been corrected to "Transaction" - two occurrences.


huong91:
#aelf #good

PR #62, opened 2018-03-16
Feature protobuf grpc

shiwk:
GRPC & ProtoBuf


shiwk:
RPC client running in UnitTest without server started up in CI, so the checks failed. The clients will be removed from UnitTest later.

PR #99, opened 2018-04-14
Feature transaction pool

shiwk:
Tx pool contains executable and waiting list (it should be sorted with IncrementId for From Accout).
Already implemented:
- add/ remove/ replace/promote/demote txs
- lock for read/write
- add Fee and TxPoolConfig

TODO:
- pool lifetime
- tx validation
- rule for Fee


AElf.Kernel/TxMemPool/TxPool.cs
```diff
@@ -0,0 +1,456 @@
...
+        
+        /// <summary>
+        /// remove unvalid txs sent by account address addr, from executable list, like too old tx
+        /// </summary>
+        /// <param name="accountHash"></param>
+        /// <returns></returns>
+        private List<Transaction> RemoveExecutedTxs(Hash accountHash)
+        {
+            var context = _accountContextService.GetAccountDataContext(accountHash, _context.ChainId);
+            var nonce = context.IncreasementId;
```
gldeng: Can use GetNonce here

PR #117, opened 2018-05-18
separate into individual classes

Justbeingjustin:
Put ResourceExtractor in a separate file


AElf.Kernel/Concurrency/ResourceExtractor.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+
+namespace AElf.Kernel.Concurrency
+{
+    public static class ResourceExtractor
```
samfrb: Hi, thanks for taking an interest in the project! Since this class is static and has no state could you maybe rename "ResourceExtractor" to "ConcurrencyHelpers" ?

PR #173, opened 2018-06-10
feat: add get commands

### LINKED ISSUE #161: RPC command "get_commands"
child of #158

Justbeingjustin:
fixes #161

### DIFF
--- a/AElf.CLI/Command/GetCommandsCmd.cs
+++ b/AElf.CLI/Command/GetCommandsCmd.cs
@@ -1,0 +1,33 @@
+﻿using AElf.CLI.Parsing;
+using System;
+using System.Text;
+
+namespace AElf.CLI.Command
+{
+    public class GetCommandsCmd : CliCommandDefinition
+    {
+        private const string GetPeersName = "getcommands";
+        
+        public GetCommandsCmd() : base(GetPeersName)
+        {
+        }
+
+        public override string GetPrintString(string resp)
+        {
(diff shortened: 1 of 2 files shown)


AElf.CLI/Command/GetCommandsCmd.cs
```diff
@@ -0,0 +1,33 @@
...
+        public GetCommandsCmd() : base(GetPeersName)
+        {
+        }
+
+        public override string GetPrintString(string resp)
+        {
+            StringBuilder strBuilder = new StringBuilder();
+            strBuilder.AppendLine("peers\t Displays peers in the network.");
+            strBuilder.AppendLine("account\t Displays account information.");
+            return strBuilder.ToString();
```
samfrb: The idea behind the "getcommands" command on the CLI is to be able to query the node for the commands it exposes through the RPC interface - at the moment it's purely for informational purposes. 
These types of "usages" should go in their respective command definition classes like "GetPeersCmd".
In terms of overridable methods : 
GetUsage() : this is the help string,
GetPrintString() : returns the string that will be the response printed on screen.
Validate() : any type of error checking needed
For the cli we're currently working on the "rpc-pipeline", if you modify you're code according to the comments and PR to that branch I'll accept the contribution :)

Justbeingjustin: Gotcha.
It appears that the overridable method "GetUsage" isn't being called anywhere. Where will that come in?

PR #178, opened 2018-06-13
Feature transaction metadata extraction

PadawanZH:
Add resource detection of a transaction: combine the metadata of the called function and the "influenced account" in the transaction


PadawanZH:
Due to change resource in ResourceUsageDetectionService from a set of hash to a set of string, some test case fail. Need to integrate the metadata and adjust the related logic in parallel execution.

PR #180, opened 2018-06-14
SDK improvements and ABI

gldeng:


gldeng:
PR is no longer needed as there is a newer PR #182

PR #253, opened 2018-07-01
Feature disable metadata

gldeng:
Metadata functionalities are disabled by default.
To enable them, compile with PARALLEL defined.


PadawanZH:
Need to change ResourceUsageDetectionService when grouper instantiate

PR #295, opened 2018-07-09
rebuild cli command

mashujie:
rebuild cli command

### DIFF
--- a/AElf.CLI/AElfCliProgram.cs
+++ b/AElf.CLI/AElfCliProgram.cs
@@ -1,16 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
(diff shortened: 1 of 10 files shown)


reyoung1110:
@mashujie Feel free to check out and take a glance of #299 

I think the `CLI.Commands` could be redesigned to make sure all logic of one command is in the same place.

Thanks.

mashujie:
@reyoung1110 I will learn your demo code suggestions #299 and improve the `CLI.Commands`. 
Thank you.

reyoung1110:
@mashujie Or if it is not in hurry, let me give a complete PR? Or let us commit to #299 together?
I just want to use AElf.CLI as the entry point to understand the code logic.

PR #375, opened 2018-08-03
fix rpc api : get_block_info and get_tx_result

mengruipan:


AElf.Node/Node/RPC/RpcServer.cs
```diff
@@ -312,7 +321,9 @@ private async Task<JObject> ProGetBlockInfo(JObject reqParams)
                     },
                     ["Body"] = new JObject
                     {
-                        ["TransactionsCount"] = blockinfo.Body.TransactionsCount
+                        ["TransactionsCount"] = blockinfo.Body.TransactionsCount,
+                        ["Transactions"] = JArray.FromObject(txs)
```
rosona: Please add parameter indicate return the transaction list.

Issue #377, opened 2018-08-04
Serialization issue for LogEvent related to DeploySmartContract


gldeng:
Hyperion serialization cannot recognize some protobuf message types.  The issue was found when we tested DeploySmartContract execution in remote worker.

The method DeploySmartContract fires a ContractHasBeenDeployed event which is logged in TransactionTrace and Hyperion has problem with serialization of such a TransactionTrace. The proposed solutions are:

- [x] Workaround: disable ContractHasBeenDeployed event in SmartContractZero so that testing can be done without any issues. #378 
- [x] Medium Term fix: Convert TransactionTrace to protobuf and keep only the deserialized TransactionTrace in actor's TransactionTraceMessage #380 
- [x] Long Term solution: Make all actor message in protobuf format (IMessage) and use protobuf serialization for all IMessage

PR #378, opened 2018-08-04
Workaround for TransactionTrace serialization issue

gldeng:
#377


gldeng:
Closing this because a better solution in #380 is done.

PR #489, opened 2018-09-20
Fix recover DPoS mining when restart single node

EanCuznaivy:


AElf.Kernel.Consensus/AElfDPoSHelper.cs
```diff
@@ -347,7 +347,7 @@ private Round GenerateNextRoundOrder()
                 foreach (var node in _miners.Nodes)
                 {
                     var s = this[node].Signature;
-                    if (s == null)
+                    if (s == null || s.GetHashBytes().Length < 30)
```
rosona: As a temporary solution, this fix should be improved.

Issue #491, opened 2018-09-21
Rpc: Current Rpc middleware caused CPU increment quickly durning much transaction request


EricBlack:
Repro steps:
1. Launch standalone node, including parallel or not parallel mode.
2. Send a few transaction requests(2~3txs/second) to node continues.

Result:
Multi CPU all quickly increment to 100%, and block cannot generate on time or cannot generated.

Screenshot:
![image](https://user-images.githubusercontent.com/5166529/45859252-e5520500-bd93-11e8-8b26-327c0f3e70f9.png)

EricBlack:
This issue is caused by long tcp connect. If too much long connect request, it would be caused cpu increment quickly.

EricBlack:
Close due to issue fixed.

PR #502, opened 2018-09-28
Set KeepAliveTimeout to smallest unit to effectively disable KeepAlive

gldeng:
Fix for #491


gldeng:
Hold the PR for now.
Give client the freedom to choose keep alive.
But the issue has to be handled in production environment.

PR #729, opened 2018-12-20
enhancement cross chain contract

shiwk:


shiwk:
- Index multi parent chain blocks at one time.
- Add transaction  for indexing side chain.

PR #733, opened 2018-12-21
Feature module and bootstrapper 20181217

loning:
use ABP to manage module and bootstrap, and maybe use module configuration to replace current configuration in AElf.

- [x] USE ABP module system
- [x] unit tests framework
- [x] move old Run method in IAElfModule to a new way like app.useXXX();
- [ ] remove the dependency of configuration when configure services.
- [ ] replace static configuration system to  ABP configuration of Module.
- [x] remove dependency of AUTOFAC in module project.
- [ ] merge all RPC/webapi servers into one server. a node when startup, it will start a MVC website. and controller will be injected. and module dependency will decide what controllers will be injected.
- [x] auto register service/manger to services collection
- [x] remove dependency of NLog in module project.
- [ ] use asp.net core mvc to implement RPC server instead of the json rpc server
- [x] change logging style


loning:
do not merge, just view file changes

loning:
I have changed the basic framework, so there are some more works need to be done.

global search 
//TODO!

loning:
`
catch (Exception e)
            {
                Logger.LogError("CreateNewChainAsync Error: " + e);
                return null;
            }
`

we should let the exception throw out

PR #740, opened 2018-12-24
Release/0.7.0

loning:
- [x] remove logging attributes
- [ ] rename AElf.Kernel.Types to AElf.Types
- [ ] remove the dependency of configuration when configure services.
- [ ] replace static configuration #800 
- [ ] merge all RPC/webapi servers into one server. a node when startup, it will start a MVC website. and controller will be injected. and module dependency will decide what controllers will be injected.
- [ ] use asp.net core mvc to implement RPC server instead of the json rpc server
- [x] startup
- [ ] AElf.Network => GRPC


loning:
- [x] Only inmemory db type testing works, run with redis and ssdb will throw exception.
- [ ] Config file not works, only command line works normally.
- [ ] Standalone node launch successful, but rpc api service not works. Call connec_chain api return "500 Internal Server Error" error.
- [ ] Run with CLI2 to get node height, return error 'Invalid JSON RPC response: {"error":"HttpRequestorError","details":"JSON.parse Error: Unexpected input at position:0"} '.
- [x] Output log print not including timestamp, it's not convenient for issue debugging, especially for some stuck situation.
- [x] Not verify multiple nodes due to cannot run on one machine with inmemory db, will test when other types db works.
- [ ] Not verify user transaction due to rpc api and CLI2 not available.

loning:
@EricBlack also if possible, bugs with more details such as exception details much better

EricBlack:
> @EricBlack also if possible, bugs with more details such as exception details much better

Ok, I will provided more details in later new bugs, thanks for your reminder. 
For your branch, do you think I just log new issues in issue list or sent comments in your PR request, which do you prefer?

loning:
@EricBlack RPC server can work now, please test

PR #766, opened 2019-01-02
Feature/new Launcher and RPC server

loning:
- [x] create a new asp.net core 2.0 project AElf.FullNodeHosting
- [x] remove TerminatedModule
- [x] register JSON-RPC service
- [x] unit tests for RPC service and RPC server

I have noticed, when we implement RPC server, we also use a IOC container.
Currently we create a new server and use a new IOC container.

In my view, we can build a full node in the stack of asp.net core, and make the RPC as a module, we can also still use the RPC implement.

and also, I have noticed that the WebManagement and RPC miner have different RPC implement, maybe we should make them in the same way.

Can we implement a service that support GRPC, JSON-RPC and WEBAPI in the same time?


loning:
@samfrb @shiwk

PR #768, opened 2019-01-02
Fix node sync up

samfrb:
- Peer.IsSyncingHistory has been reverted to its original meaning. 
- The blocks are, for now, not requested in parallel. 
- Issue to be re-implemented: during normal sync, a failed request needs to trigger a sync to someone else.
- Fix init sync when target is lower than first announce (sync up to the lowest announce height).


samfrb:
fixes #768

PR #793, opened 2019-01-08
Feature shared bp

shiwk:
- [x] Fix incorrect block execution result.
- [x] Enable side chain indexing fee.
- [x] Fix irregular LIBs.
- [x] Complete shared BP cross chain.


shiwk:
Fix #796

PR #814, opened 2019-01-14
Feature/statemanager 20190113

loning:
- [ ] add inMemory cache for BlockStateSet


loning:
how to use: https://github.com/AElfProject/AElf/pull/814/files#diff-0e63505a2142e99e732703ed425299d3

@gldeng

PR #821, opened 2019-01-16
Improve protobuf types & dependency

rosona:
- [ ] Move all protobuf file to AElf.Types project and update the referenced project
- [ ] Rebuild types dependency and refactor types partial class
- [ ] Review/improve protobuf types, namespace and classification


loning:
ok, but we may change the ci shell script to generate protobuf cs files

rosona:
> https://github.com/travis-ci/container-example/blob/master/install-protobuf.sh

In the beginning, this solution(shell script) was used, there is no problem with CI, and there are several inconveniences for developers:
- It maybe inconveniences if not auto generate cs files with dotnet build, it needs to be run separately.
- The shell script has platform compatibility (windows)

There is another solution I have seen before: https://grpc.io/blog/grpc-dotnet-build, Do you think it is better? 

Today I will merge this PR first, I will include this changes in the next PR.

PR #838, opened 2019-01-24
Refactor/network

samfrb:
Tasks related to the refactor, these can change according to remarks and redesign:

- [x] Add proto definitions and wire up the csproj.
- [x] Init: connect to bootnodes.
- [x] Report handshake logic.
- [x] Transaction dispatch.
- [ ] Report other less important items like timeouts, keepalive...
- [x] Move to OS.
- [ ] Cleanup, review and tests.
- [x] Rewire RPC get/add/remove peers.


AElf.OS.Network.Grpc/GrpcNetworkManager.cs
```diff
@@ -0,0 +1,275 @@
...
+            Logger = NullLogger<GrpcNetworkManager>.Instance;
+            
+             _localEventBus = localEventBus;
+            
+            _authenticatedPeers = new List<GrpcPeer>();
+            
+            _networkOptions = options.Value;
+        }
+        
+        public async Task StartAsync()
```
loning: a manager should not have status, so we need to remove server start out of manager

we should make the server as singleton, and start it in the OnApplicationInitialize method in the module of the project

AElf.OS/Network/INetworkManager.cs
```diff
@@ -0,0 +1,20 @@
...
+    public interface INetworkManager
+    {
+        Task<bool> AddPeerAsync(string address);
+        Task<bool> RemovePeerAsync(string address);
+        List<string> GetPeers();
+        
+        Task<IBlock> GetBlockByHash(Hash hash, string peer = null);
+        Task BroadcastAnnounce(Block b);
+        
+        Task StartAsync();
```
loning: remove start and stop

AElf.OS/Network/INetworkManager.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace AElf.OS.Network
+{
+    public interface INetworkManager
+    {
+        Task<bool> AddPeerAsync(string address);
+        Task<bool> RemovePeerAsync(string address);
+        List<string> GetPeers();
+        
+        Task<IBlock> GetBlockByHash(Hash hash, string peer = null);
+        Task BroadcastAnnounce(Block b);
```
loning: do not broadcast a block, just only send the announce of {block hash, height} or block header

AElf.OS/Network/INetworkService.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using AElf.Common;
+using AElf.Kernel;
+
+namespace AElf.OS.Network
+{
+    public interface INetworkService
+    {
+        void AddPeer(string address);
```
loning: if network service has the same method with network manager, we do not need it.

just check the different between manager and service

AElf.OS.Network.Grpc/GrpcNetworkManager.cs
samfrb: I agree, and in this case I propose we rename the class and not use a "Manager" for the network layer (I don't see why we should use one actually (?)). This class (I'm not sure what to name it yet) can of course be Singleton, no problem.
Also ok for using OnApplicationInit for starting the net layer.

AElf.OS/Network/INetworkManager.cs
samfrb: see comment about "GrpcNetworkManager", it maybe just needs renaming.

AElf.OS/Network/INetworkManager.cs
samfrb: Yes agreed, no need for the block at this point.

AElf.OS/Network/INetworkService.cs
samfrb: In this case, as said in other comments, we don't really need anything that's a "Manager". In my opinion we can use the "NetworkService" (used internally by our RPC and handlers in our system) and "ServerSingleton" (maybe not final name), these two are enough.

AElf.OS.Network.Grpc/GrpcPeerService.cs
```diff
@@ -0,0 +1,191 @@
...
+        /// clients authentication information. When receiving this call, protocol dictates you send the client your auth
+        /// information. The response says whether or not you can connect.
+        /// </summary>
+        public override Task<AuthResponse> Connect(Handshake request, ServerCallContext context)
+        {
+            Logger?.LogTrace($"[{context.Peer}] has initiated a connection request.");
+            
+            try
+            {
+                var peerServer = context.Peer.Split(":")[1] + ":" + request.HskData.ListeningPort;
```
loning: context.Peer.Split(":")[1]

may make a class, Parse(string ) ToString().

do not split everywhere...

loning: or you can store it using Url class

AElf.OS.Network.Grpc/GrpcNetworkService.cs
```diff
@@ -0,0 +1,110 @@
...
+        
+        private IAElfNetworkServer _server;
+
+        public GrpcNetworkService(IAElfNetworkServer server)
+        {
+            _server = server;
+            _logger = NullLogger<GrpcNetworkServer>.Instance;
+        }
+
+        public void AddPeer(string address)
```
loning: => Task AddPeerAsync

AElf.OS.Network.Grpc/GrpcNetworkService.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            _server.AddPeerAsync(address);
+        }
+
+        public Task RemovePeer(string address)
```
loning: RemovePeerAsync

loning: also change other method names

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -0,0 +1,207 @@
...
+            Logger = NullLogger<GrpcNetworkServer>.Instance;
+            EventBus = NullLocalEventBus.Instance;
+            
+            _authenticatedPeers = new List<GrpcPeer>();
+        }
+        
+        public async Task StartAsync()
+        {
+            // todo inject block service
+            var p = new GrpcPeerService(this, _blockService);
```
loning: maybe you should implement IPeerAuthentificator in another class ,

you can implement a factory to avoid use new operation like https://github.com/abpframework/abp/blob/55458a7a081f6eb6124d55950d4d61e5a48f5fc7/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/IocEventHandlerFactory.cs

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -0,0 +1,207 @@
...
+            p.EventBus = EventBus;
+            
+            p.PeerSentDisconnection += POnPeerSentDisconnection;
+            
+            _server = new Server {
+                Services = { PeerService.BindService(p) },
+                Ports = { new ServerPort(IPAddress.Any.ToString(), _networkOptions.ListeningPort, ServerCredentials.Insecure) }
+            };
+            
+            _server.Start();
```
loning: should use await Task.Run(_server.Start);

cause _server.Start has some IO operations may block the async method.

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -0,0 +1,207 @@
...
+            
+            // Add the provided boot nodes
+            if (_networkOptions.BootNodes != null && _networkOptions.BootNodes.Any())
+            {
+                foreach (var btn in _networkOptions.BootNodes)
+                {
+                    await Dial(btn);
```
loning: maybe Task.WaitAll(_networkOptions.BootNodes.Select(btn => Dial(btn)).ToArray())

loning: if one timeout 2s, 10 boots we need to wait 20s, but parallel connection can keep wait only 2s.

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -0,0 +1,207 @@
...
+                Logger.LogError(e, $"Error while connection to {address}.");
+                return false;
+            }
+        }
+
+        public async Task StopAsync()
+        {
+            await _server.KillAsync();
+            
+            foreach (var peer in _authenticatedPeers)
```
loning: maybe we do not need to tell other peer we are offline.

other node will do ping to us, it should have a logic to keep alive.

samfrb: In my mind order is:
1. Disconnect message when we can (clean disconnect, clear connection asap, can even help debugging if we had a "reason")
2. Keepalive, this will eventually detect that the peer is offline (grpc does this).

loning: if we need a restart work, send shutdown message to other nodes may make the shutdown time much longer..

maybe we can leave these, and if in the future we need them, we add the feature that time.

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -58,15 +58,13 @@ public async Task StartAsync()
...
+            await Task.Run(() => _server.Start());
             
             // Add the provided boot nodes
             if (_networkOptions.BootNodes != null && _networkOptions.BootNodes.Any())
             {
-                foreach (var btn in _networkOptions.BootNodes)
-                {
-                    await Dial(btn);
-                }
+                List<Task<bool>> taskList = _networkOptions.BootNodes.Select(Dial).ToList();
```
loning: var tasks = _networkOptions.BootNodes.Select(Dial).As<Task>().ToArray();
Task.WaitAll(tasks);

AElf.OS.Network.Grpc/GrpcNetworkServer.cs
```diff
@@ -94,7 +92,7 @@ private async Task<bool> Dial(string address)
                 var client = new PeerService.PeerServiceClient(channel);
                 var hsk = BuildHandshake();
                         
-                var resp = await client.ConnectAsync(hsk);
+                var resp = await client.ConnectAsync(hsk, new CallOptions().WithDeadline(DateTime.UtcNow.AddSeconds(2)));
```
loning: inject a options class to replace 2

PR #920, opened 2019-02-23
Added a few handlers

gldeng:


AElf.OS.Network.Grpc/GrpcUrl.cs
```diff
@@ -17,6 +17,11 @@ public static GrpcUrl Parse(string address)
             return url;
         }
         
+        public string ToIpPortFormat()
```
loning: You can use System.Uri or System.Net.IpAddress

no need to implement by your self

loning: @samfrb

AElf.OS.Network.Grpc/GrpcPeerPool.cs
```diff
@@ -134,7 +134,7 @@ public GrpcPeer FindPeer(string peerEndpoint, byte[] publicKey = null)
             IEnumerable<GrpcPeer> toFind = _authenticatedPeers;
 
             if (!string.IsNullOrWhiteSpace(peerEndpoint))
-                toFind = toFind.Where(p => p.PeerAddress == peerEndpoint);
+                toFind = toFind.Where(p => p.PeerAddress == peerEndpoint || peerEndpoint.EndsWith(p.RemoteListenPort));
```
loning: do not use port to match peer.

just public key

AElf.OS.Network.Grpc/GrpcUrl.cs
samfrb: will be solved by #940

AElf.OS.Network.Grpc/GrpcPeerPool.cs
samfrb: will be solved by #940

PR #922, opened 2019-02-25
improve unit test

zhxymh:
AElf.Kernel
  - [x] BlockGenerationService
  - [x] FullBlockchainService
  - [x] TransactionResultManager
  - [x] ChainCreationService
  - [x] FullBlockchainExecutingService


AElf.Kernel.Tests/Managers/ChainManagerTests.cs
```diff
@@ -1,573 +0,0 @@
-using System;
```
loning: why deleted？

zhxymh: ChainManagerTests is already exists. somebody add a new one, and commented this before.

PR #969, opened 2019-02-27
[WIP] Add default keystore path (content root) for better devops

rosona:


rosona:
Less reasonable demand

PR #970, opened 2019-02-27
Refactor txhub

gldeng:
- [x] Accepts Transaction only if refblock is after LIB, store it once received


loning:
transaction.ExpiryBlockNumber maybe we can store it in zero contract,or just consider it as const

every transaction has ExpiryBlockNumber may make tx more larger,

AElf.Contracts.TestBase/ContractTester.cs
```diff
@@ -303,7 +317,7 @@ private MinerService BuildMinerService(List<Transaction> txs, List<Transaction>
 
             return new MinerService(mockTxHub.Object, _accountService, _blockGenerationService,
```
loning: miner service should not depend on a tx Hub

maybe minerService has miner, miner can have a tx hub

gldeng: TxHub needs to be retrieved from IChainRelatedComponentManager<ITxHub>, can MinerService depend on IChainRelatedComponentManager<ITxHub>?

PR #996, opened 2019-02-27
add a test base to initialize database config and other common configs

loning:


loning:
@zhxymh @EricBlack 
when you copy codes, make sure remove duplicated code, 
for example, you can use base class, also depends on a module

zhxymh:
> @zhxymh @EricBlack
> when you copy codes, make sure remove duplicated code,
> for example, you can use base class, also depends on a module

ok

PR #1053, opened 2019-03-04
make serialize / deserialize handle null

loning:


loning:
@shiwk @EanCuznaivy some contracts cannot pass

PR #1064, opened 2019-03-05
Change system transaction generator interface.

### LINKED ISSUE #1063: Transaction generation interface.
For this system txn generation [interface](https://github.com/AElfProject/AElf/blob/dev/AElf.Kernel.Core/Miner/Application/ISystemTransactionGenerator.cs#L8-L9), only previous block hash **prefix** is available but not real block hash. It can be changed to whole hash data? Because I'm using readonly txn to read state which needs the previous block hash as chain context. 

BTW, **Consensus** module uses **IBlockChainService** to get current information about the **chain** to make it work now.

shiwk:
Fix #1063


shiwk:
Fix #1063

PR #1084, opened 2019-03-06
NewIrreversibleBlockFoundEvent and continuous LIB handling in TransactionResultService

gldeng:


AElf.Kernel.Core/Blockchain/Application/IBlockchainService.cs
```diff
@@ -5,6 +5,7 @@
 using AElf.Common;
 using AElf.Kernel.Blockchain.Domain;
 using AElf.Kernel.Blockchain.Events;
+using AElf.Kernel.SmartContract.Domain;
```
loning: do not introduce this namespace.

SmartContract reference Blockchain

PR #1118, opened 2019-03-09
smart contract runtime

loning:


loning:
@gldeng I have moved the sdk context into Kernel.SmartContext to let the SmartContract connect with   the host by the context.

maybe we should only notify the smart contract the context has been changed such as the transaction context has been changed.

and also want to unify the state's context, the smart contact's context in the future

Issue #1156, opened 2019-03-13
'codecov' cannot spot the unit cases covered correctly about Contract test cases. 


EricBlack:
## Description
 
With 'codecov' tool to check currently test cases coverage, for Contracts test cases cannot distinguished out correctly. 
Select Token contract project for example:
- codecov report screenshot for View method
![image](https://user-images.githubusercontent.com/5166529/54270237-ed0f9b00-45b9-11e9-9076-e5acba19170e.png)
- code covered cases
```
[Fact]
        public async Task Initialize_View_TokenContract()
        {
            await Initialize_TokenContract();

            var bytes1 = await Tester.CallContractMethodAsync(TokenContractAddress, nameof(TokenContract.TotalSupply));
            bytes1.DeserializeToUInt64().ShouldBe(1000_000UL);
            var bytes2 = await Tester.CallContractMethodAsync(TokenContractAddress, nameof(TokenContract.Decimals));
            bytes2.DeserializeToUInt64().ShouldBe(2U);
            var byte3 = await Tester.CallContractMethodAsync(TokenContractAddress, nameof(TokenContract.TokenName));
            byte3.DeserializeToString().ShouldBe("elf token");
            var byte4 = await Tester.CallContractMethodAsync(TokenContractAddress, nameof(TokenContract.Symbol));
            byte4.DeserializeToString().ShouldBe("ELF");
        }
```
- For Token contract Action methods
![image](https://user-images.githubusercontent.com/5166529/54269866-21cf2280-45b9-11e9-83ff-50e951c670b0.png)
- code covered cases
```
[Fact]
        public async Task UnApprove_TokenContract()
        {
            await Approve_TokenContract();
            var owner = Tester.GetCallOwnerAddress();
            var spender = Tester.GetAddress(_spenderKeyPair);

            var result2 =
                await Tester.ExecuteContractWithMiningAsync(TokenContractAddress, nameof(TokenContract.UnApprove),
                    spender, 1000UL);
            result2.Status.ShouldBe(TransactionResultStatus.Mined);
            var bytes2 = await Tester.CallContractMethodAsync(TokenContractAddress, nameof(TokenContract.Allowance),
                owner, spender);
            bytes2.DeserializeToUInt64().ShouldBe(2000UL - 1000UL);
        }

        [Fact]
        public async Task UnApprove_Without_Enough_Allowance()
        {
            await Initialize_TokenContract();

            var owner = Tester.GetCallOwnerAddress();
[...]

PR #1192, opened 2019-03-13
code coverage test

loning:


loning:
fix https://github.com/AElfProject/AElf/issues/1156

PR #1194, opened 2019-03-13
Split cli related project to  AElf.Console

llyyss1988:
- remove cli related project from aelf.sln to aelf.console
- support macos, linux, windows for contract.csharp.plugin


AElf.CLI.sln
```diff
@@ -0,0 +1,46 @@
+﻿
```
rosona: Remove AElf.CLI and AElf.CLI.Tests from AElf.sln

PR #1212, opened 2019-03-15
Use MultiToken Contract instead of old Token Contract.

EanCuznaivy:


loning:
if we reference in Contracts.TestBase, we do not need to ref it in other tests project

maybe we should define a AElf.Kernel.Token project with TokenService, to avoid directly construct the ELF token transaction @EanCuznaivy

PR #1237, opened 2019-03-16
Lock and Unlock token in MultiToken Contract.

EanCuznaivy:


loning:
@EanCuznaivy need to resolve conflict

PR #1313, opened 2019-03-20
Change Interface for filling block.

shiwk:


AElf.Kernel.Core/Blockchain/Application/BlockExtraDataService.cs
```diff
@@ -39,8 +40,35 @@ public ByteString GetExtraDataFromBlockHeader(string blockExtraDataProviderSymbo
                     return blockHeader.BlockExtraDatas[i];
                 }
             }
-
             return null;
         }
+
+        public void FillMktRootExtraDataForTransactionStatus(BlockHeader blockHeader,
```
loning: Mkt

shiwk: > Mkt

Fixed in #1317 .

PR #1333, opened 2019-03-21
Enhancement/clean branch after set LIB

zhxymh:


AElf.Kernel.Core/Blockchain/Application/IBlockchainService.cs
```diff
@@ -248,6 +248,45 @@ public async Task SetBestChainAsync(Chain chain, long bestChainHeight, Hash best
                 BlockHeight = irreversibleBlockHeight
             };
             await _chainManager.SetIrreversibleBlockAsync(chain, irreversibleBlockHash);
+
+            // Clean branches and not linked
+            var bestChainKey = chain.BestChainHash.ToStorageKey();
```
loning: move these codes into chainManager

1. chainManager.CleanBranches(Hash libBlockHash, long libHeight) returns List<Hash>
2. remove blocks, transaction in blockchain service.

zhxymh: 1. In this logic, need to use method GetBlockHashByHeightAsync, can I move the method GetBlockHashByHeightAsync to manager too?
2. I think we can't remove transaction, because one transaction maybe in two blocks in different branch.

AElf.Kernel.Core/Blockchain/Domain/IChainManager.cs
```diff
@@ -289,5 +296,87 @@ public int GetChainId()
...
+            {
+                chain.NotLinkedBlocks.Remove(key);
+            }
+
+            await _chains.SetAsync(chain.Id.ToStorageKey(), chain);
+
+            return toRemoveBlocks;
+        }
+
+        private async Task<(List<Hash>, List<string>, List<string>)> GetToCleanBlocksAndBranchesAsync(Chain chain,
```
loning: no return this (List<Hash>, List<string>, List<string>)

PR #1334, opened 2019-03-21
handle not covered unit cases

EricBlack:


AElf.Sdk.CSharp.Tests/EventTests.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    public class EventTests
+    {
+        // Lot of types data should not be defined including: byte, sbyte, char, short, ushort, decimal, float and double.
+        [Fact]
+        public void Parse_EventTest()
+        {
+            var eventDemo = new EventDemo()
+            {
+               //init index value
```
loning: make them in different events and then check them

EricBlack: have updated code and test with different events.

AElf.Sdk.CSharp.Tests/EventCollection.cs
```diff
@@ -0,0 +1,67 @@
...
+        [Indexed] public long IndexLong {get;set;}
+
+        public Address NonIndexAddress {get;set;}
+        public string NonIndexString {get;set;}
+        public int NonIndexInt {get;set;}
+    }
+
+    public class NotSupportEvent : Event
+    {
+        [Indexed] public double IndexDouble {get;set;}
```
loning: should define in different classes, each only contains one 
 public class NotSupportDoubleIndexEvent : Event
    {
        [Indexed] public double IndexDouble {get;set;}
}
 public class NotSupportDoubleEvent : Event
    {
        public double IndexDouble {get;set;}
}

EricBlack: updated not support event type case one by one.

PR #1347, opened 2019-03-21
remove old runtime

loning:


loning:
@gldeng @EanCuznaivy we clean up in the PR

PR #1349, opened 2019-03-22
Fix state merge issue

rosona:


loning:
@rosona should not return a key-value, or tuple, I have create an interface

PR #1425, opened 2019-03-26
resolve test case #1418

EricBlack:


AElf.Kernel.Core.Tests/KernelCoreTestAElfModule.cs
```diff
@@ -69,6 +71,11 @@ public override void ConfigureServices(ServiceConfigurationContext context)
                    
                     return dataProvider.Object;
                 });
+            
+            //For BlockchainStateMergingService testing
```
loning: remove

EricBlack: @loning thx boss for code review, learned a lot~

PR #1473, opened 2019-04-01
Add web api unit test

jason-hoopox:


loning:
@jason-hoopox net need to be tested

PR #1479, opened 2019-04-02
Performance - execution

zhxymh:


AElf.Kernel.SmartContractExecution.Benches/BlockExecutingTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        private Block _block;
+        
+        private Counter _counter;
+        
+        public BlockExecutingTests(ITestOutputHelper output)
+        {
+            Trace.Listeners.Clear();
+            Trace.Listeners.Add(new XunitTraceListener(output));
+            
+            _blockchainService = GetRequiredService<IBlockchainService>();
```
loning: move to setup

zhxymh: ok

PR #1545, opened 2019-04-13
 Improve benchmark

zhxymh:
- [x] Support for testing using databases
- [x] Fix clean up problem
- [x] Export redis database test results
- [ ] Find problems with performance test results


zhxymh:
The problems with performance test results will be fixed in other pr.

PR #1613, opened 2019-04-23
Remove random number in contract 

EricBlack:


src/AElf.Contracts.TestContract.Basic1Update/Basic1Contract_Action.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+using AElf.Sdk.CSharp;
 using Google.Protobuf.WellKnownTypes;
 
 namespace AElf.Contracts.TestContract.Basic11
```
loning: no XXX1,YYY2

EricBlack: have updated project name, will follow naming rule in future.

PR #1615, opened 2019-04-23
Fix contract update

### LINKED ISSUE #1580: After updating contract, the method of contract is not changed
After updating contract check the method of contract, it was not changed. 
and update contract again with same files. the transaction is failed and the error is AElf.Sdk.CSharp.AssertionException: Code is not changed.

zhxymh:
Fix #1580


zhxymh:
> if you have running executive, it was already taken out the pool, then you upgrade, you will get the hell.
> 
> 1. a=GetExecutiveAsync
> 2. UpdateCache
> 3. b=GetExecutiveAsync
> 4. PutExecutiveAsync(a)
> 
> what will your cache be? the old b executive will be in the pool
> 
> so we have to compare reg's hash and the executive's hash
> 
> And we should not pass SmartContractRegistration in the interface,
> 
> just add a method called ClearExecutivePool(Address address)

On question one, I discussed with GuangLei before. This situation happens in parallel execution. In GuangLei's opinion, our execution needs to avoid this situation. That is to say, for a contract, update and method call are conflicting, and get and put executive are not allowed when the contract is upgrading, so no need to compare the hash.

On question two, I will change the method to ClearExecutivePool.
@loning

loning:
Service is stateless，and should work in parallel condition

发自我的 iPhone

zhxymh:
> Service is stateless，and should work in parallel condition 发自我的 iPhone
> […](#)
> 在 2019年4月25日，10:30，ZhangXin ***@***.***> 写道： if you have running executive, it was already taken out the pool, then you upgrade, you will get the hell. a=GetExecutiveAsync UpdateCache b=GetExecutiveAsync PutExecutiveAsync(a) what will your cache be? the old b executive will be in the pool so we have to compare reg's hash and the executive's hash And we should not pass SmartContractRegistration in the interface, just add a method called ClearExecutivePool(Address address) On question one, I discussed with GuangLei before. This situation happens in parallel execution. In GuangLei's opinion, our execution needs to avoid this situation. That is to say, for a contract, update and method call are conflicting, and get and put executive are not allowed when the contract is upgrading, so no need to compare the hash. On question two, I will change the method to ClearExecutivePool. @loning — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

Yes. But in parallel condition, we need to place the transaction that updates the contract in a special place. If we treat it as a normal transaction, then we can't get the same order of execution at all nodes, so we will get different results. 
For example, if a block contains a transaction that call contract A and a transaction that updates contract A, the order of execution of the two transactions will be different on different nodes, and the execution result will be different.

loning:
Don’t care about how other people use, make your service methods can work in parallel condition 

发自我的 iPhone

zhxymh:
> Don’t care about how other people use, make your service methods can work in parallel condition 发自我的 iPhone
> […](#)
> 在 2019年4月25日，11:00，ZhangXin ***@***.***> 写道： Service is stateless，and should work in parallel condition 发自我的 iPhone … 在 2019年4月25日，10:30，ZhangXin @.***> 写道： if you have running executive, it was already taken out the pool, then you upgrade, you will get the hell. a=GetExecutiveAsync UpdateCache b=GetExecutiveAsync PutExecutiveAsync(a) what will your cache be? the old b executive will be in the pool so we have to compare reg's hash and the executive's hash And we should not pass SmartContractRegistration in the interface, just add a method called ClearExecutivePool(Address address) On question one, I discussed with GuangLei before. This situation happens in parallel execution. In GuangLei's opinion, our execution needs to avoid this situation. That is to say, for a contract, update and method call are conflicting, and get and put executive are not allowed when the contract is upgrading, so no need to compare the hash. On question two, I will change the method to ClearExecutivePool. @loning — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread. Yes. But in parallel condition, we need to place the transaction that updates the contract in a special place. If we treat it as a normal transaction, then we can't get the same order of execution at all nodes, so we will get different results. For example, if a block contains a transaction that call contract A and a transaction that updates contract A, the order of execution of the two transactions will be different on different nodes, and the execution result will be different. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

Ok, I will add the ContractCodeHash attribute to the Executive. Meanwhile, the types of ExecutivePools need to be changed from <Address, ConcurrentBag<IExecutive>> to <Hash, ConcurrentBag<IExecutive>>. Because according to the previous type, PutExecutive needs to check the Hash first and then put it back into the cache, it also have concurrency problems.

zhxymh:
What's your opinion? @loning

loning:
@zhxymh seems not good,
let me do the left job, you may solve other issues

zhxymh:
> @zhxymh seems not good,
> let me do the left job, you may solve other issues

ok,thx.

PR #1617, opened 2019-04-23
Change address to 32 bytes

rosona:


test/AElf.Types.Tests/AddressTest.cs
```diff
@@ -20,7 +20,7 @@ public void Generate_Address()
             address3.ShouldNotBe(null);
 
             //Generate from byte
-            var bytes = new byte[30];
+            var bytes = new byte[32];
             new Random().NextBytes(bytes);
```
loning: should not use random
should throw exception

rosona: > should not use random
> should throw exception

Fixed:
1. remove TakeByAddressLength function
2. remove random gen bytes
3. check throw exception when bytes length < 32
4. check address length (gen from bytes > 32)

PR #1618, opened 2019-04-23
update setup instructions

loning:
- Clarified some text.
- Fixed directory path.
- Restructured as follows:
  1. Basic dependencies like Git, Dotnet core...
  2. Install script for additional libraries like protobuf.
  3. Build and run.


loning:
@samfrb @hzz780 please add required install instructions

PR #1624, opened 2019-04-25
Implement contract security audit structure

aerkol:


aerkol:
Structure is more or less done now, need to enhance the rules based on the existing system contracts.

aerkol:
Non-whitelisted libraries are referred from contracts because of their assemblies are loaded during run time and they contain injected codes for code coverage. Need to load the DLL directly after the build.

src/AElf.Runtime.CSharp.Core/Validators/Method/MultiDimArrayValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                if (instruction.OpCode.Code == Code.Newobj)
+                {
+                    var methodRef = (MethodReference) instruction.Operand;
+
+                    if (methodRef.DeclaringType.IsArray && ((ArrayType) methodRef.DeclaringType).Dimensions.Count > 1)
+                    {
+                        return new List<ValidationResult>
+                        {
+                            new MultiDimArrayValidationResult(
```
gldeng: This is inconsistent with adding all results in output like in UnsafeMathValidator

src/AElf.Runtime.CSharp.Core/Validators/Whitelist/Whitelist.cs
```diff
@@ -0,0 +1,246 @@
...
+            foreach (var asmRef in module.AssemblyReferences)
+            {
+                if (!_assemblies.Keys.Contains(asmRef.FullName))
+                    results.Add(new WhitelistValidationResult("Assembly " + asmRef.FullName + " is not allowed."));
+            }
+            
+            // Validate types in the module
+            foreach (var typeDef in module.Types)
+            {
+                results.AddRange(Validate(typeDef));
```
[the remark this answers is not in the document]
aerkol: Done

src/AElf.Runtime.CSharp.Core/Validators/Method/MultiDimArrayValidator.cs
aerkol: Done

contracts/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.dll
```
rosona: Can I delete this folder if just for testing only?

aerkol: Now I moved "contracts" folder to AElf.Runtime.CSharp.Tests project. The reason why I copied the DLLs to a separate folder at build time was to prevent codecov injection into contract DLLs. When the DLLs from its output directory used, they are auto injected with codecov tracking methods and causing the test case to fail.

PR #1628, opened 2019-04-25
Fix benchmark startup

zhxymh:
BenchmarkSwitcher had some problems with the default config, and I'll troubleshooting that later.


loning:
have u test under release mode？

zhxymh:
> have u test under release mode？

Yes, it works.

PR #1634, opened 2019-04-28
Improve doc about js demo

hzz780:


docs/tutorials/dapp/setup.md
```diff
@@ -112,6 +112,18 @@ npm start
 
 You should see the results in the terminal or in the browser dev tool.
 
+### In the terminal
+
+<p align="center">
+  <img src="assets/jssdk_terminal.jpg" width="400">
+</p>
```
rosona: try this format?  avatar

loning: png or jpg is ok

loning: maybe you just need to copy the output, and put into 
```
OUTPUT
```

hzz780: eeee, a good idea.
But image may be more intuitive.

Issue #1636, opened 2019-04-28
Improve block/transaction check


rosona:
- [x] Move MerkleTreeRoot verify to block save function
- [x] Move BlockValidation to blocksync（before attach logic）
- [x] Remove transactionList from blackbody

rosona:
Fixed

PR #1644, opened 2019-04-29
remove the full transaction list from the BlockBody

rosona:
Fix issue #1636 

- Removing the full transaction list from the _BlockBody_ type.

**Description:**
The original intent for this type was for storing the block without the full transactions (these are in another store), so it’s confusing and error prone to keep the _TransactionList_ field in this type.

**Solution:**
Add another type used when we need the full transactions like when answering a block request from another peer.

**Todo:**
- [x] use BlockWithTransaction on the network layer (send/recv)
- [x] convert smart contract GetBlockWithTransaction instead of normal (for getting the senders of previous block)
- [x] recover tests.

Important changes:
* Added a proto type to replace the TransactionList in the BlockBody.
* Network now uses this type to send blocks.
* Chain creation now uses a BlockWithTransactions.
* The attach service now as a dedicated method to receive (attach) a BlockWithTransactions.
* System transactions are saved directly after being signed by the node.


src/AElf.Kernel.Types/Block/BlockWithTransactions.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// Used to override IMessage's default string representation.
+        /// </summary>
+        /// <returns></returns>
+        public string ToDiagnosticString()
+        {
+            return $"{{ id: {GetHash()}, height: {Height} }}";
+        }
+
+        public BlockHeader Header
```
rosona: Try init header with BlockWithTransactions creation (remove blockheader getter/setter)

src/AElf.Kernel.Types/Block/BlockWithTransactions.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            return Header.GetHashBytes();
+        }
+
+        Block IBlock.Clone()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Block ToBlock()
```
rosona: Move ToBlock to extensions method?

samfrb: why not :)

src/AElf.Kernel.Types/Block/IBlock.cs
```diff
@@ -2,10 +2,9 @@
 {
     public interface IBlock : IBlockBase
     {
-        BlockHeader Header { get; set; }
-        BlockBody Body { get; set; }
-        long Height { get; set; }
+        BlockHeader Header { get; }
+        BlockBody Body { get; }
+        long Height { get; }
```
loning: remove height

samfrb: Remove and use the Header property to access height ? Some methods currently use IBlock.Height

src/AElf.Kernel.SmartContractExecution/Application/IBlockAttachService.cs
```diff
@@ -27,24 +28,48 @@ public class BlockAttachService : IBlockAttachService, ITransientDependency
...
+            
+            if (!await ValidateAsync(block))
+                return;
+            
+            await _blockchainService.AddBlockWithTransactionsAsync(blockWithTransactions);
+            await AttachAndExecute(block);
+        }
+
         public async Task AttachBlockAsync(Block block)
+        {
```
rosona: Transactions in block produced by himself will be lost (cannot be saved)

src/AElf.Kernel.SmartContractExecution/Application/IBlockAttachService.cs
```diff
@@ -9,6 +9,7 @@ namespace AElf.Kernel.SmartContractExecution.Application
     public interface IBlockAttachService
     {
         Task AttachBlockAsync(Block block);
+        Task AttachReceivedBlock(BlockWithTransactions blockWithTransactions);
```
gldeng: When should AttachReceivedBlock be used and when should AttachBlockAsync be used?
It seems AttachBlockAsync is only used after mining, then it should not include the logics for executing again.

samfrb: Correct, AttachReceivedBlock is used for blocks received from the network and AttachBlockAsync is intended for mined blocks.

This is an interesting comment, I largely reused the behavior from dev. Currently, on dev, the mined block will be executed twice. I'm not sure we can remove without refactoring because the _AttachBlockToChainAsync_ method will modify the state of the _ChainBlockLinks_ and this is further updated by executing. If you simply remove execution, it will leave the node in an incorrect state after mining.

In the end it seems that the operations after mining can be simplified.

src/AElf.Contracts.TestKit/TransactionExecutor.cs
```diff
@@ -30,7 +30,10 @@ public async Task ExecuteAsync(Transaction transaction)
                 new List<Transaction> {transaction},
                 DateTime.UtcNow, TimeSpan.FromMilliseconds(int.MaxValue));
 
-            await blockAttachService.AttachBlockAsync(block);
+            var blckWithTrx = new BlockWithTransactions {Header = block.Header};
```
shiwk: naming style, Change BlockWithTransactions -> BlockHeaderWithTransactions

src/AElf.Kernel.Core/Blockchain/Application/IBlockchainService.cs
```diff
@@ -65,12 +68,12 @@ public static class BlockchainServiceExtensions
             return await blockchainService.GetBlockByHashAsync(hash);
         }
 
-        public static async Task<List<Block>> GetBlocksInBestChainBranchAsync(this IBlockchainService blockchainService,
+        public static async Task<List<BlockWithTransactions>> GetBlocksInBestChainBranchAsync(this IBlockchainService blockchainService,
             Hash firstHash,
```
shiwk: Move get block with tx to network extensions method.

src/AElf.Kernel.SmartContractExecution/Application/IBlockAttachService.cs
```diff
@@ -27,24 +28,48 @@ public class BlockAttachService : IBlockAttachService, ITransientDependency
...
 
+        public async Task AttachReceivedBlock(BlockWithTransactions blockWithTransactions)
+        {
+            Block block = blockWithTransactions.ToBlock();
+            
+            if (!await ValidateAsync(block))
+                return;
+            
+            await _blockchainService.AddBlockWithTransactionsAsync(blockWithTransactions);
+            await AttachAndExecute(block);
```
shiwk: Save txs in network while receive block from network

src/AElf.Kernel.Types/Block/Block.cs
```diff
@@ -25,6 +26,8 @@ public Block(Hash preBlockHash)
             Body = new BlockBody();
         }
 
+        public IEnumerable<Hash> TransactionList => Body.Transactions;
```
shiwk: rename to TransactionHashList

src/AElf.Kernel.Types/Block/IBlock.cs
shiwk: Use header's height

src/AElf.Kernel/Miner/Application/MinerService.cs
```diff
@@ -71,18 +72,23 @@ public class MiningService : IMiningService
         private readonly IBlockGenerationService _blockGenerationService;
         private readonly IAccountService _accountService;
         private readonly IBlockExecutingService _blockExecutingService;
+        private readonly ITransactionManager _transactionManager;
```
shiwk: Inject BlockchainService to insert transaction

src/AElf.OS/Jobs/BlockSyncJob.cs
```diff
@@ -113,7 +113,7 @@ public async Task ExecuteAsync(BlockSyncJobArgs args)
                         }
 
                         Logger.LogDebug($"Processing block {block},  longest chain hash: {chain.LongestChainHash}, best chain hash : {chain.BestChainHash}");
-                        _taskQueueManager.Enqueue(async () => await _blockAttachService.AttachBlockAsync(block),
+                        _taskQueueManager.Enqueue(async () => await _blockAttachService.AttachReceivedBlock(block),
```
shiwk: naming "*Async"

src/AElf.Kernel.Types/Block/Block.cs
samfrb: for the TODO, opened a tracking issue : #1692

src/AElf.Kernel.SmartContractExecution/Application/IBlockAttachService.cs
samfrb: you suggest saving from the BlockSyncJob ?

PR #1676, opened 2019-05-07
web api validate transaction for send transaction

jason-hoopox:


test/AElf.WebApp.Application.Chain.Tests/BlockChainAppServiceTest.cs
```diff
@@ -132,6 +133,83 @@ public async Task Broadcast_Transaction_ReturnInvalidTransaction()
 
             response.Error.Code.ShouldBe(Error.InvalidTransaction.ToString());
             response.Error.Message.ShouldBe(Error.Message[Error.InvalidTransaction]);
+            
+            var contractAddress = _smartContractAddressService.GetAddressByContractName(TokenSmartContractAddressNameProvider.Name);
+
+            var from = Base64.ToBase64String("InvalidAddress".ToUtf8Bytes());
```
rosona: don't use RedisProtocol's extensions/helper (will cleanup later)

src/AElf.WebApp.Application.Chain/BlockChainAppService.cs
```diff
@@ -665,5 +667,17 @@ private async Task<MethodDescriptor> GetContractMethodDescriptorAsync(Address co
 
             return null;
         }
+        
+        private void ValidateMessage(IMessage message,int errorCode)
```
rosona: return true/false is better for "ValidateMessage", the user can decide whether to throw an exception or logging error/warn

PR #1677, opened 2019-05-07
Fix high memory usage problem

zhxymh:


loning:
add a unit test to ensure the cache work

zhxymh:
> add a unit test to ensure the cache work

ok

PR #1679, opened 2019-05-08
Remove the longest branch that cannot be executed

zhxymh:


src/AElf.Kernel.SmartContractExecution/Application/IBlockExecutingService.cs
```diff
@@ -115,6 +120,12 @@ await LocalEventBus.PublishAsync(new BlockAcceptedEvent()
             {
                 Logger.LogWarning($"Block validate fails after execution. Exception message {ex.Message}");
             }
+            catch (Exception e)
```
rosona: Keep the same naming style for exception

src/AElf.Kernel.Core/Blockchain/Domain/IChainManager.cs
```diff
@@ -410,5 +411,16 @@ private async Task RemoveChainBranchesAsync(Chain chain, List<string> branchKeys
 
             await _chains.SetAsync(chain.Id.ToStorageKey(), chain);
         }
+
+        public async Task RemoveCannotExecutedLongestBranchAsync(Chain chain)
```
loning: Can you name it in a OOP way ???

zhxymh: How about "RemoveNonExecutableLongestBranchAsync"?

zhxymh: > Can you name it in a OOP way ???

Sorry, I got it. Maybe "RemoveLongestBranchAsync" is better?

PR #1684, opened 2019-05-09
Move files

jason-hoopox:


bench/AElf.Benchmark/MinerTests.cs
```diff
@@ -6,11 +6,11 @@
 using AElf.Kernel.Blockchain.Application;
 using AElf.Kernel.Blockchain.Domain;
 using AElf.Kernel.Blockchain.Events;
-using AElf.Kernel.EventMessages;
 using AElf.Kernel.Infrastructure;
 using AElf.Kernel.Miner.Application;
 using AElf.Kernel.SmartContract.Infrastructure;
 using AElf.Kernel.SmartContractExecution.Application;
+using AElf.Kernel.SmartContractExecution.Events;
```
loning: move to AElf.Kernel.SmartContractExecution.Application,

service and event should put in the same namespace

src/AElf.Types/Extensions/HashExtensions.cs
```diff
@@ -8,5 +8,22 @@ public static Hash ComputeHash(this int obj)
         {
             return Hash.FromRawBytes(BitConverter.GetBytes(obj));
         }
+        
+        // TODO: Consider Span
```
rosona: Try implement this TODO

src/AElf.Kernel.Consensus.DPoS/Application/ConsensusService.cs
```diff
@@ -2,8 +2,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using AElf.Kernel.Consensus.Events;
```
EanCuznaivy: No need to add namespace xx.Events.

PR #1701, opened 2019-05-14
Contract security enhancement, limit array size and element types.

aerkol:
- [x] Limit array size and element types by total memory consumption or by count if runtime memory consumption of element type is not known.
- [x] Add test cases for malicious scenarios.


test/AElf.Runtime.CSharp.Tests/ContractAuditorTests.cs
```diff
@@ -157,6 +166,13 @@ public void CheckBadContract_ForDeniedMemberUseInSeparateClass()
                 .ShouldNotBeNull();
         }
         
+        [Fact]
+        public void CheckBadContract_ForLargeArrayInitialization()
+        {
+            _findings.FirstOrDefault(f => f.GetType() == typeof(ArrayValidationResult) && f.Info.ReferencingMethod == "InitLargeArray")
```
gldeng: May be we can use `f is ArrayValidationResult`

aerkol: Fixed by the latest commit

PR #1704, opened 2019-05-15
Fix package build error

rosona:


rosona:
> I will later change these to follow the same approach as Guanglei did in his current PR.

ok

PR #1706, opened 2019-05-15
Use context to get contract address

jason-hoopox:


loning:
State.ProfitContractSystemName.Value 

do not need to read from state, directly use Hash.FromString("AElf.ContractNames.XXXX")

PR #1711, opened 2019-05-16
Fix bp node sometimes can't set lib

jason-hoopox:


src/AElf.OS.Consensus.DPos/DPoSAnnouncementReceivedEventDataHandler.cs
```diff
@@ -90,7 +94,8 @@ public async Task<IBlockIndex> FindLastLastIrreversibleBlockAsync(string senderP
 
             var peers = _peerPool.GetPeers().Where(p => pubkeyList.Contains(p.PubKey)).ToList();
 
-            if (peers.Count == 0)
+            var pubKey = await _accountService.GetPublicKeyAsync();
+            if (peers.Count == 0 && !pubkeyList.Contains(pubKey.ToHex()))
```
rosona: Full node' pub key is not in pubkeylist, so the full node cannot get LIB from network.

jason-hoopox: If peers count is greater than 0, this condition will be false.

PR #1713, opened 2019-05-16
Fix reconnection problem

### LINKED ISSUE #1690: Fix avoidable auth error during the connection process
**Problem description**

Currently, when one node (N1) connects to another (N2) it will create a 2 way connection. In practice this means that two gRPC channel are created: (N1 -> N2) and (N2 -> N1). 
During the connection process, there's an authentification phase, that will reject the connection - based on the pub-key - if this peer is already in the peer pool.

For example, when the channel (N1->N2) is created N2 will verify that he does not already have N1 as a peer. If for whatever reason, N2 still has an old gRPC channel to N1 (sometimes a perfectly functioning because gRPC will do most of the underlying reconnection work) N2 will consider that there's a problem and will reject the connection.

There is no real reason why we would not try and recover from this situation.

**Proposal**

This can be fixed with the following: when the peer is connecting and we already have a channel to that peer, shut it down, clean up and create another.

samfrb:
Closes #1690


src/AElf.OS.Network.Grpc/GrpcServerService.cs
```diff
@@ -69,22 +69,34 @@ public override async Task<ConnectReply> Connect(Handshake handshake, ServerCall
...
+            
+            var old = _peerPool.FindPeerByPublicKey(pubKey);
 
+            if (old != null)
+            {
+                Logger.LogDebug($"Cleaning up {old} before connecting.");
+                await _peerPool.RemovePeerAsync(pubKey, false);
+            }
+
+            var peerAddress = peer.IpAddress + ":" + handshake.HskData.ListeningPort;
```
rosona: Uri for this?

src/AElf.OS.Network.Grpc/GrpcServerService.cs
```diff
@@ -69,22 +69,34 @@ public override async Task<ConnectReply> Connect(Handshake handshake, ServerCall
...
             var peer = GrpcUrl.Parse(context.Peer);
-            if(peer == null)
+            
+            if (peer == null)
                 return new ConnectReply { Err = AuthError.InvalidPeer };
             
-            var peerAddress = peer.IpAddress + ":" + handshake.HskData.ListeningPort;
+            var pubKey = handshake.HskData.PublicKey.ToHex();
+            
+            var old = _peerPool.FindPeerByPublicKey(pubKey);
```
samfrb: oldPeer

PR #1714, opened 2019-05-16
Fix contracts that contain double type and float operations

aerkol:
- [x] Add TimestampExtensions class to be used in contracts for time operations.
- [x] Replace time operations and other codes in contracts that cause double type usage and float operations in IL codes.
- [x] Fix remaining audit errors in contracts
- [x] Enable contract auditor (replace AssemblyChecker with the new one)
- [x] Use TimestampHelper when getting current time.


src/AElf.Contracts.Election/ElectionContract_Election.cs
```diff
@@ -209,14 +209,14 @@ public override Empty Vote(VoteMinerInput input)
...
-            var lockDays = (input.EndTimestamp - Context.CurrentBlockTime.ToTimestamp()).ToTimeSpan().TotalDays;
+            var lockDays = (input.EndTimestamp - Context.CurrentBlockTime.ToTimestamp()).Days();
             Assert(lockDays >= State.MinimumLockTime.Value,
                 $"Invalid lock time. At least {State.MinimumLockTime.Value} {(TimeUnit) State.BaseTimeUnit.Value}");
             Assert(lockDays <= State.MaximumLockTime.Value,
                 $"Invalid lock time. At most {State.MaximumLockTime.Value} {(TimeUnit) State.BaseTimeUnit.Value}");
 
             State.LockTimeMap[Context.TransactionId] =
-                GetTimeSpan(input.EndTimestamp.ToDateTime(), Context.CurrentBlockTime);
+                GetTimeSpan(input.EndTimestamp, Context.CurrentBlockTime.ToTimestamp());
```
loning: directly change the context interface

aerkol: To confirm, I will change Context.CurrentBlockTime type from DateTime to Timestamp?

aerkol:
Building contracts with overflow check helps passing the contract auditor validations, however it somehow breaks some of the test cases that shows the signs that there is an overflow during the tests. Disabling overflow check now and not enabling build with overflow check in contracts. This is to be fixed later.

aerkol:
**Summary**
- We need to implement IL code processing monitoring to prevent DOS completely.
- We need to fix failing test cases when contracts are built with "check for arithmetic overflow", then we can enable float ops validator in contract auditor. Currently all contracts passes this check once they are built with -checked flag but unfortunately test cases fail. This points out to an existing overflow in contracts with those test cases, further investigation is required.
- Preventing float op codes brings some minor disadvantages such as contract devs can't use foreach to iterate through arrays (ienumerables are fine).

src/AElf.Contracts.AssociationAuth/AElf.Contracts.AssociationAuth.csproj
```diff
@@ -20,4 +20,8 @@
     <ItemGroup>
       <ProjectReference Include="..\AElf.Sdk.CSharp\AElf.Sdk.CSharp.csproj" />
     </ItemGroup>
+    <Target Name="CopyContractDll" AfterTargets="AfterBuild">
```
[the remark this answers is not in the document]
aerkol: Added the copy task to current targets file

bench/AElf.Benchmark/Fibonacci16Tests.cs
```diff
@@ -55,7 +55,7 @@ public async Task Fibonacci16()
                     BlockHeight = _chain.BestChainHeight
                 },
                 _transaction,
-                DateTime.UtcNow);
+                DateTime.UtcNow.ToTimestamp());
```
loning: change to Helper method, do not directly use DateTime.UtcNow

contract/AElf.Contracts.Consensus.AEDPoS/Types/MinerList.cs
```diff
@@ -32,7 +33,7 @@ public partial class MinerList
                 minerInRound.PublicKey = sortedMiners[i];
                 minerInRound.Order = i + 1;
                 minerInRound.ExpectedMiningTime =
```
loning: do not need change TimeStamp to DateTime, directly use TimeStamp + duration

src/AElf.Sdk.CSharp/SafeDateTime.cs
```diff
@@ -0,0 +1,146 @@
+using System;
+using Google.Protobuf.WellKnownTypes;
+
+namespace AElf.Sdk.CSharp
+{
+    public class SafeDateTime
```
loning: remove safe date time, only let smart contract use Timestamp and Duration, because DateTime has DateTime.Kind and double

contract/AElf.Contracts.Consensus.AEDPoS/Types/MinerList.cs
aerkol: Should I add extensions for Timestamp? Like AddMilliseconds, AddSeconds then use duration to add inside the extensions?

loning:
we can merge and change it later

aerkol:
> we can merge and change it later

Sure, adding DateTimeHelper now, will be pushing soon

bench/AElf.Benchmark/BlockExecutingTests.cs
```diff
@@ -51,7 +52,7 @@ public async Task IterationSetup()
                     ChainId = chain.Id,
                     Height = chain.BestChainHeight + 1,
                     PreviousBlockHash = chain.BestChainHash,
-                    Time = Timestamp.FromDateTime(DateTime.UtcNow)
+                    Time = DateTimeHelper.Now.ToTimestamp()
```
loning: TimestampHelper.Now, I remembered some one has an implement

aerkol: Sure, I will use that one

aerkol:
Fixed conflicts now fixing the remaining build errors

PR #1716, opened 2019-05-17
Send announcement to multiple peers at the same time.

samfrb:
- Task.WhenAll for sending.
- broadcast announce is now fire and forget.

### DIFF
--- a/src/AElf.OS.Core/Network/Application/NetworkService.cs
+++ b/src/AElf.OS.Core/Network/Application/NetworkService.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/AElf.OS.Core/Network/Application/NetworkService.cs
```diff
@@ -46,28 +47,31 @@ public List<IPeer> GetPeers()
...
+                    Interlocked.Increment(ref successfulBcasts);
                 }
                 catch (NetworkException e)
                 {
                     Logger.LogError(e, "Error while sending block.");
                 }
-            }
-
+            }));
+
```
loning: var tasks = _peerPool.GetPeers().Select(async .....).ToList();
await Task.WhenAll(tasks);
var success = tasks.Count(t=>t.Status == Task.Success);

PR #1725, opened 2019-05-20
ACS2 MetaDataProvider

loning:


loning:
@gldeng I have implemented a very simple ACS2, to return the resource of a method

gldeng:
> @gldeng I have implemented a very simple ACS2, to return the resource of a method

I understand the idea already. Will implement full functionality.

PR #1747, opened 2019-05-27
Feature local parallel execution

gldeng:


src/AElf.Kernel.Core/Blockchain/Events/TransactionResourceExtractionEvents.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+using AElf.Types;
+
+namespace AElf.Kernel.Blockchain.Events
+{
+    public class TransactionResourcesNeededEvent
```
gldeng: I think it's better we use more generic event types instead of creating specific events. Core shouldn't have knowledge about parallel execution (which is an extension).

aerkol: I pushed a new one to use existing events and remove the new ones specific to resource extraction.

PR #1766, opened 2019-06-01
Fix contract reference problem

zhxymh:


bench/AElf.Benchmark/Program.cs
```diff
@@ -8,6 +11,7 @@ class Program
     {
         static void Main(string[] args)
         {
+            AddSameFolderLoader();
```
loning: rename the function，so chinglish

PR #1769, opened 2019-06-03
Queue the broadcast tx work on the pool.

samfrb:
To avoid causing useless timeouts and retries on the client side (broadcaster) the server needs to handle the broadcast as soon as possible. For broadcasting transactions, the client only needs to “transmit” the data and should not have to wait for any processing. 

Using Task.Run (lightweight version of Task.Factory.StartNew) will queue the work on the thread pool and finish the RPC request as soon as possible.


samfrb:
Closing this because a better solution is implemented in #1793

PR #1770, opened 2019-06-03
Added case for connection bug.

samfrb:
During test and heavy load/bad connections it can happen that the client has a timeout, but the server sometimes has time to add the client as a peer. Meaning that when the server tries to communicate with the client, the client rejects the call on the basis that the handshake has not being done.

The workaround here is for this special case, but should eventually be generalized to other network problems.


src/AElf.OS.Network.Grpc/GrpcPeer.cs
```diff
@@ -250,17 +250,26 @@ private void HandleFailure(AggregateException exceptions, string errorMessage)
...
+            {
+                _ = StopAndRemove();
+            }
             else
             {
                 throw new NetworkException($"Failed request to {this}: {errorMessage}", exceptions);
             }
         }
 
+        private async Task StopAndRemove()
```
loning: StopAndRemoveAsync  

Async method should end with Async,

I hope it's my last comment of a name of function

src/AElf.OS.Network.Grpc/GrpcPeer.cs
```diff
@@ -250,17 +250,26 @@ private void HandleFailure(AggregateException exceptions, string errorMessage)
...
+        {
+            await StopAsync();
+            DisconnectionEvent?.Invoke(this, EventArgs.Empty);
```
loning: what's this ?

samfrb: This method is used to shutdown the peer, the DisconnectionEvent is subscribed by the peer pool, when this is fired the peer pool will remove this peer from its internal list of peers.
So the event is fired in 2 cases for now:
1. The situation that this PR fixes (see description).
2. When the service uses a peer to request or broadcast and that the underlying grpc channel is not healthy (TRANSIENT_FAILURE, it can happen, maybe the channel is bad for just 1 or 2 seconds), the logic will wait a short time for the channel to recover, if after the wait it is still not recovered it will call StopAndRemoveAsync.

loning: we should not use C# event

just use EventBus

samfrb: ok, updated

src/AElf.OS.Network.Grpc/GrpcServerService.cs
```diff
@@ -132,8 +132,8 @@ private GrpcPeer DialPeer(string peerAddress, Handshake handshake)
                 StartHeight = handshake.Header.Height,
                 IsInbound = true
             };
-
-            return new GrpcPeer(channel, client, connectionInfo);
+
```
loning: do not pass EventBus

samfrb: So can we put the peer in the DI system (create with GetService<IPeer>()) ? this way the logger will be injected.
I can also see another solution, we can throw an exception here (maybe PeerDisconnectedException) and move the logic to the NetworkService. This way GrpcPeer does not need the EventBus.

samfrb: Moved the logic for handling failing peers from GrpcPeer to the NetworkService.

PR #1778, opened 2019-06-11
Stop mining if one miner detected he is alone.

EanCuznaivy:


loning:
stop mine until one node connect to other，and the  total number of nodes is not 1 in config files

发自我的 iPhone

PR #1793, opened 2019-06-15
[WIP] Improvements for the network

samfrb:
* TxHub: move verify signature after checking that we already have transaction.
* Broadcast transactions and announcements: both will use a queue (different queues) with a specified amount of workers (options available).
* Receive transactions and announcements: enqueued in their respective queues, one worker per queue.
* Avoid useless broadcasts.


samfrb:
Closing this because #1845 already fixes the network related improvements and the signature check in TxHub is already on dev.

PR #1808, opened 2019-06-20
Broadcast immediately after mining

zhxymh:


src/AElf.Kernel/BlockMinedEventData.cs
```diff
@@ -0,0 +1,9 @@
+namespace AElf.Kernel
```
rosona: Move BlockMinedEventData to Kernel.Core/Blockchain/Events?

PR #1813, opened 2019-06-21
Increase the miners' count when the term changed;

YanhuaViolet:


contract/AElf.Contracts.Consensus.AEDPoS/AEDPoSContract.cs
```diff
@@ -181,17 +184,15 @@ public override Empty NextRound(Round input)
...
-                var minersCount = GetMinersCount();
+                var minersCount = GetMinersCount(input);
                 if (minersCount != 0 && State.ElectionContract.Value != null)
                 {
                     State.ElectionContract.UpdateMinersCount.Send(new UpdateMinersCountInput
                     {
                         MinersCount = minersCount
                     });
                 }
+                Context.LogDebug(()=>"the count of the round:"+minersCount);
```
rosona: The logging is not clear: "the count of the round"? Remove the logging if not used. BTW: It's best to use $"" format string.

PR #1829, opened 2019-06-23
 Limit the number of parameters

stillwindows7:
Limit the number of parameters  ，Keeping the number of parameters low makes units easier to understand, test and reuse.


rosona:
Need to split the service reasonably, please reference #1789: “ split them into different AppServices”

PR #1847, opened 2019-06-26
Break continuous mining.

EanCuznaivy:


contract/AElf.Contracts.Consensus.AEDPoS/Types/Round_Generation.cs
```diff
@@ -3,6 +3,7 @@
 using System.Linq;
 using AElf.Sdk.CSharp;
 using Google.Protobuf.WellKnownTypes;
+using Org.BouncyCastle.Asn1.Cms;
```
loning: which need this?

EanCuznaivy: > which need this?

Referenced by mistake.

PR #1878, opened 2019-07-04
Fix repeated injection of BlockValidationProvider

zhxymh:
BlockValidationProvider is injected three times, then block is validated three times.

- When ABP loads the modules, the assembly has been registered, so should not do it again.
- When the assembly is registered, ABP is going to go through all ConventionalRegistrar and register.

So remove all AddAssemblyOf, and modify the logic of custom ConventionalRegistrar.


src/AElf.Core/AElfDefaultConventionalRegistrar.cs
```diff
@@ -7,25 +8,27 @@ namespace AElf
...
+        private const string _interfacePrefix = "I";
+
+        private readonly List<string> _transientTypeSuffixes = 
+            new List<string>{"Service", "Provider", "Manager", "Store", "Factory"};
+
         protected override ServiceLifetime? GetServiceLifetimeFromClassHierarcy(Type type)
         {
             var lifeTime = base.GetServiceLifetimeFromClassHierarcy(type);
-            if (lifeTime != null)
+            if (lifeTime == null)
```
loning: in your implements, `var lifeTime = base.GetServiceLifetimeFromClassHierarcy(type);` did nothing, remove it will get the same result

if you do not know the details of how these work, please do not change it.

What's the meanning of `private const string _interfacePrefix = "I";`? Will you use other instead of using "I"?

src/AElf.Core/Modularity/AElfModule.cs
```diff
@@ -10,9 +9,5 @@ public abstract class AElfModule : AbpModule
     public abstract class AElfModule<TSelf> : AElfModule
         where TSelf : AElfModule<TSelf>
     {
-        public override void ConfigureServices(ServiceConfigurationContext context)
```
loning: ????

loning:
@shiwk @rosona I hope you can understand every codes, or please do not approve it

PR #1881, opened 2019-07-05
Fix failing to set transaction limit after restart node

jason-hoopox:


src/AElf.Core/AElfDefaultConventionalRegistrar.cs
```diff
@@ -12,7 +12,7 @@ public class AElfDefaultConventionalRegistrar : DefaultConventionalRegistrar
             var lifeTime = base.GetServiceLifetimeFromClassHierarcy(type);
             if (lifeTime != null)
             {
-                return lifeTime;
+                return null;
```
loning: please do not change the implements,

PR #1897, opened 2019-07-10
Refactoring to the network design.

samfrb:
**Goals of this refactor**

1. Layer separation:
- **OS**: handlers for application events and workers.
- **OS.Core**: 
  - Application: exposes functionality to higher layers (OS handlers, other projects...)
  - Infrastructure: defines the needs of OS.Core's application layer -> (IAElfServer, IPeer) and a container for storing current peers.
- **OS.Network.Grpc**: listening for connections, message IO.. Anything else needed to implement OS.Core's infrastructure layer.

2. OOP principles:
Overall well respected, a part from the peer pool that did not respect the **Single Responsibility** principle: 
  - handling connections has been moved to the server
  - handshake has been moved to a **HandshakeProvider** (protocol level handshake).
  - the recent block cache has been moved to its own class.


src/AElf.OS.Network.Grpc/Helpers/ConnectionHelpers.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using AElf.Kernel;
+
+namespace AElf.OS.Network.Grpc
+{
+    public static class ConnectionHelpers
```
rosona: rename helpers to Extensions

src/AElf.OS.Network.Grpc/GrpcNetworkModule.cs
```diff
@@ -1,22 +1,59 @@
...
-            context.Services.AddSingleton<PeerService.PeerServiceBase, GrpcServerService>();
-            
+            // Internal dependencies
+            context.Services.AddTransient<IPeerDialer, PeerDialer>();
+            context.Services.AddSingleton<GrpcServerService>();
             context.Services.AddSingleton<AuthInterceptor>();
             context.Services.AddSingleton<RetryInterceptor>();
+
+            // setup the server
+            context.Services.AddSingleton<Server>(o =>
```
rosona: Why not init server in GrpcNetworkServer

samfrb: Initially it was in GrpcNetworkServer, but it created extra dependencies for nothing (the AuthInterceptor) that actually are more related to the set up of the module, based on the options. If an interceptor is optional it will be never be register in the DI and will never be a dependency.

loning:
Keep Unit Interfaces Small failed, maybe we should improve some codes

PR #1901, opened 2019-07-11
Fix failing to find registration from zero contract error for group transaction

jason-hoopox:


src/AElf.Kernel.SmartContract.Parallel/TransactionGrouper.cs
```diff
@@ -32,13 +32,13 @@ public class TransactionGrouper : ITransactionGrouper, ISingletonDependency
             Logger = NullLogger<TransactionGrouper>.Instance;
         }
 
-        public async Task<(List<List<Transaction>>, List<Transaction>)> GroupAsync(List<Transaction> transactions)
+        public async Task<(List<List<Transaction>>, List<Transaction>)> GroupAsync(BlockHeader blockHeader,List<Transaction> transactions)
         {
-            var chainContext = await GetChainContextAsync();
-            if (chainContext == null)
+            var chainContext = new ChainContext()
```
loning: why not directly pass ChainContext in this function?

src/AElf.Kernel.SmartContract.Parallel/ITransactionGrouper.cs
```diff
@@ -6,6 +6,7 @@ namespace AElf.Kernel.SmartContract.Parallel
 {
     public interface ITransactionGrouper
     {
-        Task<(List<List<Transaction>>, List<Transaction>)> GroupAsync(List<Transaction> transactions);
+        Task<(List<List<Transaction>>, List<Transaction>)> GroupAsync(BlockHeader blockHeader,
```
loning: do not return Task<(List<List<Transaction>>, List<Transaction>)>,
I can't know what you are returning

aerkol: Replaced tuple with GroupedTransactions object

PR #1913, opened 2019-07-22
[WIP]Enable lock-resource

xiangxingxing:


rosona:
Discuss with team, we already charge resource token in contract, decide not to lock resource token while create side chain. so close this PR temporary.

PR #1917, opened 2019-07-25
[WIP] back-pressure handling

samfrb:
Main points:
- [x] if one of the connections is slow, it should not affect other peers.
- [ ] control memory usage for each peer, if one peer is too slow, avoid out of memory exceptions (the peers buffer should be full before the server RAM...).
- [ ] when one peer is full either drop further items or close the connection (tbd).


samfrb:
Closing, implemented differently by #1914.

PR #1918, opened 2019-07-27
Fix merkle tree

xiangxingxing:


test/AElf.Types.Tests/BinaryMerkleTestTest.cs
```diff
@@ -0,0 +1,301 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace AElf.Types.Tests
+{
+    public class BinaryMerkleTestTest
```
aerkol: Would it be better if we tested the number of nodes (count) is correct after generating a Merkle Tree. Like, from 4 leafs, 7 nodes; from 5 leafs, 13 nodes?

protobuf/aelf/core.proto
```diff
@@ -89,3 +87,18 @@ message SInt64Value
...
+}
+
+message MerklePathNode{
+    Hash hash = 1;
+    bool isLeftChildNode = 2;
+}
+
+message BinaryMerkleTree {
+    repeated Hash nodes = 1;
+    Hash root = 2;
```
aerkol: The root is already the last element in nodes, do we need to duplicate it in message? Why don't we delete it (root) from proto declaration add it in partial implementation (BinaryMerkleTree.cs) instead?

`public Hash Root => Nodes.Any() ? Nodes.Last() : Hash.Empty;`

shiwk:
@xiangxingxing Please check whether it is possible to define `BinaryMerkleTree` in a plain class, not generated by `protobuf`.

protobuf/aelf/core.proto
xiangxingxing: We should make sure all the merkle tree nodes have been computed completely , in which case the expression above is valid.

PR #1919, opened 2019-07-27
Fix execute with wrong code after updating contract

jason-hoopox:


src/AElf.Kernel.SmartContract/AElf.Kernel.SmartContract.csproj
```diff
@@ -8,6 +8,7 @@
     <PackageReference Include="Sandwych.QuickGraph.Core" Version="1.0.0" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\contract\AElf.Contracts.Genesis\AElf.Contracts.Genesis.csproj" />
```
loning: should not directly reference contract

PR #1983, opened 2019-08-27
[WIP] Improve chain security

zhxymh:
- Verify the size of the transaction in the block from the network
- Verify duplicates transaction in the block from the network
- Limit the count of blocks sent

### DIFF
--- a/src/AElf.Kernel.Core/Blockchain/Application/IBlockValidationProvider.cs
+++ b/src/AElf.Kernel.Core/Blockchain/Application/IBlockValidationProvider.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.Serialization;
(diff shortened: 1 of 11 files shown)


src/AElf.Kernel.Core/Blockchain/Application/IBlockValidationProvider.cs
```diff
@@ -94,13 +96,20 @@ public async Task<bool> ValidateBeforeAttachAsync(IBlock block)
...
             }
 
             if (block.Body.TransactionsCount == 0)
             {
-                Logger.LogWarning($"Block transactions is empty");
+                Logger.LogWarning("Block transactions is empty");
+                return false;
+            }
+            
+            var hashSet = new HashSet<Hash>();
```
rosona: use TransactionIds.Distinct().Count() != TransactionIds.count is fast?

PR #1985, opened 2019-08-29
[WIP] Side Chain Economic System.

EanCuznaivy:
Basically, the token contract will collect resource tokens from the side chain contract executing and distribute resource tokens every time when the miner list changed.


EanCuznaivy:
Contained by #2005

PR #1990, opened 2019-09-02
Improve WebApi execution time

xiangxingxing:


src/AElf.WebApp.Application.Chain/TransactionResultAppService.cs
```diff
@@ -136,9 +136,9 @@ public async Task<TransactionResultDto> GetTransactionResultAsync(string transac
             {
                 limit = Math.Min(limit, block.Body.TransactionIds.Count - offset);
                 var transactionIds = block.Body.TransactionIds.ToList().GetRange(offset, limit);
-                foreach (var hash in transactionIds)
+                foreach (var transactionId in transactionIds)
                 {
-                    var transactionResult = await GetTransactionResultAsync(hash);
+                    var transactionResult = await GetTransactionResultAsync(transactionId, block.GetHash());
```
jason-hoopox: Use realBlockHash instead of block.GetHash()

PR #1999, opened 2019-09-06
Check parameters in contracts

xiangxingxing:


contract/AElf.Contracts.MultiToken/TokenContractConstants.cs
```diff
@@ -5,5 +5,10 @@ namespace AElf.Contracts.MultiToken
     public class TokenContractConstants
     {
         public static readonly List<string> ResourceTokenSymbols = new List<string>{"RAM", "CPU", "NET", "STO"};
+        public const int WhiteListCountLimit = 50;
+        public const int TokenNameLength = 500;
+        public const int InstructionLength = 500;
```
rosona: rename (memo)

PR #2013, opened 2019-09-17
Add contract patcher and solve contract deployment timeout

aerkol:
- [x] Use AElf extensions instead of StringBuilder when appending log messages. (#1757)
- [x] Limit string size when performing string concatenation. (ContractPatcher: Replace string.Concat with AElf's Concat in contracts.)
- [x] Enforce contract deployment via creating proposal and perform contract check / patch outside of transaction execution to prevent deployment timeout (#2141)
- [x] Add test cases to test transaction generator and CodeCheckRequired event handler.
- [x] CLI tool to patch contracts. (Injection etc.) (Changed Contracts.Deployer to CLI)
If a project is a contract `<IsContract>true</IsContract>`, then it is patched.
- [x] Perform contract patching after build. (Call command from AElf.Contract.Tools.targets after build.)


rosona:
Merge this PR, create a new PR soon with the following:
- *sNameProvider should have suitable location
- SurfaceUpError need improve

PR #2016, opened 2019-09-18
Optimize consensus and economic related contracts.

### LINKED ISSUE #2000: Potential Vulnerability in Random Number Generation
The user is allowed to set `MinimumBlockHeight` where the random number will be available. Earlier requests to get random number before `MinimumBlockHeight` will fail. This is fine.

However, regardless of `MinimumBlockHeight` parameter, random number is always generated from the first or the earliest round that satisfies 2/3+1 rule from the non-null `previousInValue` of each miner in that round. (Because, random number generation is done based on `TargetRoundNumber` and it remains same although `MinimumBlockHeight` is greater than `ExpectedBlockHeight`)

`information.ExpectedBlockHeight = Math.Max(input.MinimumBlockHeight, information.ExpectedBlockHeight);`

So the requestor or anybody who has access to `RandomNumberInformationMap` (simply if they run a node) will have the ability to know or calculate the random number before `MinimumBlockHeight`. As a result, the random number will be known before it is available although a user wants the random number to be known and available at a much later block.

This is a problem.

EanCuznaivy:
The branch is based on branch `feature/mining-time-arrangement`.

- [x] Removed parameter of `RequestRandomNumber` of acs6 to fix #2000 .

- [x] Removed AElfConsensusBehaviour.UpdateValueWithoutPreivousInValue.

- [x] Add resource token symbol list to Context.Variables.


protobuf/election_contract.proto
```diff
@@ -166,6 +166,7 @@ message ElectorVote {
     repeated ElectionVotingRecord active_voting_records = 5;
     repeated ElectionVotingRecord withdrawn_votes_records = 6;
     bytes pubkey = 7;
+    aelf.Address address = 8;
```
rosona: Remove address.

PR #2065, opened 2019-09-28
[WIP] Refactor ACS1.

EanCuznaivy:
close: #2076, fix: #1974


protobuf/election_contract.proto
```diff
@@ -164,6 +164,7 @@ message ElectorVote {
     repeated ElectionVotingRecord active_voting_records = 5;
     repeated ElectionVotingRecord withdrawn_votes_records = 6;
     bytes pubkey = 7;
+    aelf.Address address = 8;
```
aerkol: Why did we add `address` when it can be generated from `pubkey`? For performance reasons?

contract/AElf.Contracts.Consensus.AEDPoS/AEDPoSContract_ViewMethods.cs
```diff
@@ -31,6 +31,11 @@ public partial class AEDPoSContract
                 }
                 : new MinerList();
 
+        public override PubkeyList GetCurrentMinerPubkeyList(Empty input) => new PubkeyList
+        {
+            Pubkeys = {GetCurrentMinerList(input).Pubkeys.Select(p => p.ToHex())}
+        };
+
```
aerkol: Why do we need another method that returns current miners' public keys as string? Now we have two methods (`GetCurrentMinerList` and `GetCurrentMinerPubkeyList`) with different names that just return the same thing in different format but it is not obvious from methods' names.

EanCuznaivy: > Why do we need another method that returns current miners' public keys as string? Now we have two methods (`GetCurrentMinerList` and `GetCurrentMinerPubkeyList`) with different names that just return the same thing in different format but it is not obvious from methods' names.

These are demands from our front devs. I'll discuss with them.

PR #2102, opened 2019-10-18
add connectors  in token convert

llypb:
1. Set connector for each resource toke


contract/AElf.Contracts.TokenConverter/BancorHelper.cs
```diff
@@ -75,7 +75,7 @@ public class BancorHelper
             if (wf == wt)
             {
                 // if both weights are the same, the formula can be reduced
-                return (long) (bf * a / (bt - a));
+                return (long) (bf / (bt - a) * a);
```
llypb: sub add

shiwk: Don't forget this.

rosona:
Improve CalculateFeeService use strategy pattern in branch: AddTxFee

PR #2136, opened 2019-10-28
Fix west time zone bug

jason-hoopox:


loning:
we can only use Timestamp，not DateTime，do not waste time and CPU on switch Time Zone

contract/AElf.Contracts.Election/ElectionContract_Maintainence.cs
```diff
@@ -53,14 +53,13 @@ public override Empty RegisterElectionVotingEvent(Empty input)
...
 
-            var startTime = DateTime.MinValue.ToUniversalTime();
             var votingRegisterInput = new VotingRegisterInput
             {
                 IsLockToken = false,
                 AcceptedCurrency = Context.Variables.NativeSymbol,
                 TotalSnapshotNumber = long.MaxValue,
-                StartTimestamp = startTime.ToTimestamp(),
-                EndTimestamp = startTime.AddTicks(DateTime.MaxValue.Ticks - DateTime.MinValue.Ticks).ToTimestamp()
+                StartTimestamp = DateTime.SpecifyKind(DateTime.MinValue,DateTimeKind.Utc).ToTimestamp(),
```
loning: StartTimestamp = Timestamp.MinValue;

ok? it is zero, no need calculation.

lpo1015:
mahaobo

PR #2148, opened 2019-11-03
Update to official BouncyCastle package

onovotny:
Portable.BouncyCastle is the official portable BouncyCastle package.


rosona:
Thanks for your PR, I added it to the verification process.

PR #2186, opened 2019-11-25
Implement execution observer to improve security

aerkol:
- [x] Implement a basic execution observer (a basic instruction counter) to inject in the beginning of each method and before each branching instruction in contracts.
Test infinite loop, test infinite recursive call 
- [x] Update contract auditor to check to ensure instruction counter was injected in contract
  - [x] Ensure `ExecutionObserverProxy` is authentic, not tampered.
  - [x] Ensure proxy `SetObserver` not called from contract.
  - [x] Ensure proxy `CallCount` is called in the beginning of each method in contract.
  - [x] Ensure proxy `BranchCount` is called before each branching opcode.
- [x] Prevent calling `MultiToken.TransferFrom` within the contract. (To prevent contracts from charging users directly.)
Done by adding a check in `Executive.ApplyAsync` when executing a transaction and disallow execution of `TransferFrom` if `CallDepth` is not 0.
- [x] Ensure ACS1 or ACS8 is implemented in the contract, by looking at the descriptors only. Not require complete implementation of methods, leave it to developer.
- [x] Tune threshold for instruction counter based on the current system contracts
- [x] Resolving test case errors, local testing


rosona:
In this PR, “TransferFrom” is hard-coded in some logic. This hard-code issue is planned to be modified in the next PR after discussion (The solution has not been determined).

PR #2261, opened 2019-12-18
Charge all resource tokens from contract address if not inefficient.

EanCuznaivy:


src/AElf.Kernel.SmartContract/Application/LocalTransactionExecutingService.cs
```diff
@@ -361,6 +362,14 @@ private static bool IsTransactionCanceled(TransactionTrace trace)
                         return false;
                     trace.PostTransactions.Add(postTx);
                     trace.PostTraces.Add(postTrace);
+
+                    if (postTx.MethodName == "ChargeResourceToken")
```
rosona: Later we will discuss this solution <postTx.MethodName == "ChargeResourceToken">. An alternative is to use event, but this solution make the transaction must have bloom log.

PR #2284, opened 2019-12-31
[WIP] Make GetHashCode deterministic

### LINKED ISSUE #2281: Contracts should not use Object method 'GetHashCode'
**Describe the bug**
The hash code of same object on different vm/machine are not the same, use hash code in contract may cause state data not consistency and caused block cannot pass verification issue. 
So contract code audit should be detected this scenario and not allow user use 'GetHashCode' in contract.

aerkol:
This PR is on hold.

- [x] Add test cases to check `GetHashCode` result across different OS
- [x] Add deterministic `GetHashCode` to AElf.Sdk
- [x] Update contract patcher to replace calls for `GetHashCode` with a deterministic one
- [x] Update default policy to prevent nondeterministic `Object.GetHashCode` calls
- [ ] Enforce overriding GetHashCode for IMessage types defined in contract
- [ ] Testing and clean up

Introduced SafeHashCode class in AElf.SDK.

After patching the contracts, all `System.Object.GetHashCode` calls goes through `SafeHashCode.GetHashCode` and it provides deterministic `GetHashCode` implementation for known types. For all unknown types, it returns `0`. We can later add support for other types if required, without re-patching or updating existing deployed contracts.

Protobuf `message` types inherited from `IMessage` types will still have their `GetHashCode` method auto generated by Protobuf and it can be used to compare objects without security risk.


aerkol:
Closing this PR since we have a new solution.

PR #2298, opened 2020-01-11
[WIP] Remove and deny GetHashCode from contracts

### LINKED ISSUE #2281: Contracts should not use Object method 'GetHashCode'

aerkol:
This PR is an attempt to fix #2281

Created this PR since the situation in #2284 is hopeless.

- [x] Remove GetHashCode implementation from contracts
- [x] Do not allow object.GetHashCode calls from contracts (Update DefaultPolicy for contract audit)
- [x] Clean up if there is any contract not passing audit
- [x] Do not allow overriding GetHashCode method in the contracts, using a type declared in contract as a key should always fall back to Equals method.
- [ ] Testing

Some improvements in #2284 for MethodCallReplacer.cs were carried to this PR.


aerkol:
Closing this PR as removing GetHashCode didn't work out.

PR #2321, opened 2020-01-16
[WIP] Validate code structure when auditing contracts

aerkol:
This PR is to enforce certain limitations in contracts:

- Do not allow any field in contract implementation, allow only const or readonly primitive types as field (for inherited types from `CSharpSmartContract`)
- For other regular types, allow only certain types as static fields (such as `Marshaller`, `MessageParser`, `FieldCodec` etc.)
- Allow only certain types such as MappedState and ContractReferenceState as fields in contract state class (for inherited types from `StateBase`)

Milestones:
- [x] Implement the main structure of the validator that identifies contract, state implementations.
- [x] Allow & deny types based on our current system contracts.
- [x] Review allowed types. (Document the risky allowed types and why they are risky to allow.)
- [ ] Reduce allowed types and update system contracts to pass the checks.
- [ ] Final documentation, reviews, testing and approval


aerkol:
Closing this PR as it is replaced with #2359

PR #2371, opened 2020-02-13
[WIP] update coverlet.msbuild and travis.yml

wudameinv:
We try to fix the problem that coverlet.msbuild fails occasionally.


rosona:
Merge to #2377

PR #2423, opened 2020-02-19
Convert network exceptions to info logs.

samfrb:
- most network exceptions are not errors, they do not represent exceptional behavior (for example timeouts are possible) and do not cause the application to malfunction.
- grpc service: avoid throwing unnecessary exception when the peer is already removed from the pool.


loning:
network is not stable，so maybe just Info，not Warn

PR #2510, opened 2020-02-26
Upgrade packages and change azure test output

loning:


loning:
close #2526, already change .net core version to 3.1.102

PR #2524, opened 2020-02-28
Add timeout for TCP connect.

samfrb:


src/AElf.OS.Network.Grpc/PeerDialer.cs
```diff
@@ -222,11 +227,18 @@ private async Task<GrpcClient> CreateClientAsync(DnsEndPoint remoteEndpoint)
...
-            
+
             TcpClient client = null;
             try
             {
-                client = new TcpClient(remoteEndpoint.Host, remoteEndpoint.Port);
+                client = new TcpClient();
+
+                if (!client.ConnectAsync(remoteEndpoint.Host, remoteEndpoint.Port)
+                    .Wait(NetworkConstants.DefaultSslCertifFetchTimeout))
```
shiwk, line 237: Try to use TaskExtensions instead of Wait.

PR #2532, opened 2020-03-02
[WIP] Improve the namespace

xiangxingxing:
Improve the namespace reviewed


test/AElf.Parallel.Tests/ParallelTestWithoutFeeAElfModule.cs
```diff
@@ -3,6 +3,7 @@
 using AElf.Kernel;
```
loning: AElf.Parallel.Tests , project name need to be changed

src/AElf.OS/Handlers/Network/BlockReceivedEventHandler.cs
```diff
@@ -11,7 +11,7 @@
 using Volo.Abp.DependencyInjection;
 using Volo.Abp.EventBus;
 
-namespace AElf.OS.Handlers
+namespace AElf.OS.Handlers.Network
```
loning: AElf.OS.Handlers.Network

what's the reference of the handler?

if it only uses network, we should put it under AElf.OS.Network, 
if it uses AElf.OS.Network and AElf.OS.xxxx, we should put it under AElf.OS

loning: AElf.OS.Handlers.Network is unacceptable

PR #2584, opened 2020-03-09
Remove fee related fields from TransactionResult.

EanCuznaivy:


src/AElf.Kernel.FeeCalculation/Application/IPrimaryTokenFeeService.cs
```diff
@@ -5,6 +5,6 @@ namespace AElf.Kernel.FeeCalculation.Application
 {
     public interface IPrimaryTokenFeeService
     {
-        Task<long> CalculateTokenFeeAsync(ITransactionContext transactionContext, IChainContext chainContext);
+        Task<long> CalculateFunction(ITransactionContext transactionContext, IChainContext chainContext);
```
rosona: why remove **Async

loning: > why remove **Async

it's in memory cache.

PR #2587, opened 2020-03-10
Improve network

zhxymh:
- [x] Remove the IsSecure of GrpcPeer, we don't need it.
- [x] Move the logic of generating certificate to TlsHelper.
- [x] Improve the logic of remove peer
- [x] Fix peer discover issue:
A peer is obtained through peer discovery, if the peer connection fails or disconnects, it can not be connected again.
- [x] Fix sync state issue:
In special cases, all nodes are restarted, and the lib of some lagging nodes cannot grow to the target lib, because the block of this lib is confirmed to have been discarded. Therefore, this part of the node needs to generate and broadcast block to form a new lib, but the broadcast block has a condition that depends on the current lib, so it forms a deadlock. So change the check condition to best chain height.

This pr contains #2559, close #2559 .


src/AElf.OS.Core/Network/Infrastructure/IBlackListedPeerProvider.cs
```diff
@@ -12,15 +13,13 @@ namespace AElf.OS.Network.Infrastructure
 {
     public interface IBlackListedPeerProvider
     {
-        bool AddHostToBlackList(string host);
+        bool AddHostToBlackList(string host, int limitSeconds);
         bool IsIpBlackListed(string host);
+        bool RemoveHostFromBlackLis(string host);
```
EanCuznaivy, line 18: RemoveHostFromBlackList

src/AElf.OS.Core/Network/Application/INetworkService.cs
```diff
@@ -8,9 +8,11 @@ namespace AElf.OS.Network.Application
 {
     public interface INetworkService
     {
-        Task<bool> AddPeerAsync(string address);
-        Task<bool> RemovePeerAsync(string address);
-        Task<bool> RemovePeerByPubkeyAsync(string peerPubKey, bool blacklistPeer = false);
+        Task<bool> AddPeerAsync(string address, bool isTrusted = false);
```
EanCuznaivy, line 11: Add interface AddTrustedPeerAsync

src/AElf.OS.Core/Network/Application/NetworkService.cs
```diff
@@ -41,25 +42,31 @@ public class NetworkService : INetworkService, ISingletonDependency
...
 
-        public async Task<bool> RemovePeerAsync(string address)
+        public async Task<bool> RemovePeerByAddressAsync(string address, int removalSeconds = NetworkConstants.DefaultPeerRemovalSeconds)
         {
             if (!AElfPeerEndpointHelper.TryParse(address, out DnsEndPoint endpoint)) 
                 return false;
             
             var peer = _peerPool.FindPeerByEndpoint(endpoint);
-            if (peer == null)
+            if (!await RemovePeerAsync(peer, removalSeconds))
```
EanCuznaivy, line 67: Add comments here

src/AElf.OS.Core/Network/Application/INetworkService.cs
```diff
@@ -8,9 +8,11 @@ namespace AElf.OS.Network.Application
...
+        Task<bool> RemovePeerByAddressAsync(string address,
```
EanCuznaivy, line 12: RemovePeerByAddressAsync ->  RemovePeerByEndPointAsync

src/AElf.OS.Core/Network/Application/NetworkService.cs
```diff
@@ -68,20 +75,27 @@ public async Task<bool> RemovePeerAsync(string address)
...
+                Logger.LogWarning($"Remove peer failed. Peer pubkey: {peerPubkey}");
                 return false;
             }
-            
-            if (blacklistPeer)
+
+            return true;
+        }
+        
+        private async Task<bool> RemovePeerAsync(IPeer peer, int removalSeconds)
```
EanCuznaivy, line 90: -> TryToRemovePeerAsync

PR #2599, opened 2020-03-18
update build.cake

wudameinv:


build.config
```diff
@@ -1,7 +1,3 @@
 #!/usr/bin/env bash
 CAKE_VERSION=0.37.0
 DOTNET_VERSION=3.1.102
-VERSION_PREFIX=0.9.3
```
loning: why do u remove build version?

build.cake
```diff
@@ -175,8 +195,25 @@ Task("Upload-Coverage-Azure")
 {
     Codecov("./CodeCoverage/Cobertura.xml","$CODECOV_TOKEN");
 });
-
+Task("Publish-Myget")
```
rosona: rename to MyGet

build.config
jason-hoopox: Build version is added in this PR.And it is set in environment variable instead of config.

PR #2624, opened 2020-03-24
Tx fee part optimize.

EanCuznaivy:


contract/AElf.Contracts.Consensus.AEDPoS/AEDPoSContract.cs
```diff
@@ -270,5 +289,31 @@ public override Hash GetRandomHash(SInt64Value input)
             Assert(Context.CurrentHeight >= input.Value, "Block height not reached.");
             return State.RandomHashes[input.Value] ?? Hash.Empty;
         }
+
+        public override Empty ContributeToSideChainDividendsPool(ContributeToSideChainDividendsPoolInput input)
```
EanCuznaivy: Add validation logic: only side chain can call this method.

PR #2716, opened 2020-04-20
Fix few bugs.

shiwk:
- [x] Fix restart problem with transaction result absent.
- [x] Fix block mining failed problem when it contains NextTerm transaction.
- [x] Fix typos.


src/AElf.Kernel/Miner/Application/IMiningRequestService.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            if (IsGenesisBlockMining(blockTime))
+                return expectedDuration;
+
+            return blockTime + expectedDuration - TimestampHelper.GetUtcNow();
+        }
+
+        private bool IsGenesisBlockMining(Timestamp blockTime)
+        {
+            return blockTime < new Timestamp {Seconds = 3600};
```
loning, line 79: why is 3600, not 0?

should make new Timestamp {Seconds = 3600} a static method, avoid new a object everytime

PR #2815, opened 2020-05-22
Little Modification

apweili:
1. Move ToByteString(this string hexString) from HexStringExtensions to ByteStringHelper.
2. Remove redundant StringExtensions.


src/AElf.Types/Helper/ByteStringHelper.cs
```diff
@@ -22,5 +22,9 @@ public static int Compare(ByteString xValue, ByteString yValue)
 
             return 0;
         }
+        public static ByteString ToByteString(this string hexString)
```
EanCuznaivy, line 25: public static ByteString FromHexString(string hexString)

PR #2854, opened 2020-06-05
[WIP] Fix EconomicTestTransactionExecutor

EanCuznaivy:


src/AElf.Contracts.TestKit/MockTxHub.cs
```diff
@@ -39,7 +40,7 @@ public async Task AddTransactionsAsync(IEnumerable<Transaction> transactions)
             var txs = transactions.ToList();
             foreach (var transaction in txs)
             {
-                _allTransactions.Add(transaction.GetHash(), transaction);
+                _allTransactions.TryAdd(transaction.GetHash(), transaction);
```
loning, line 43: if(TryAdd){
}else{
throw new OperationException("");
}

if we have any failed Add Operation, we should know it

loning: @jason-hoopox And alos check all Try Operation in aelf codes

Issue #3343, opened 2022-12-15
Verifying signature support in smart contract environment; add them in SDK.


LiWang-555:
**Is your feature request related to a problem? Please describe.**
Some smart contract need parsing json and verifying signature support, so we need to add them in SDK both.

**Describe the solution you'd like**
Just add them both.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #3346, opened 2022-12-26
Adjusted parallel resource for delegation and added verifying signature support in SDK

### LINKED ISSUE #3342: When a delegatee was involved when charging method fee, there was a potential risk of race condition on accessing the same delegatee in parallel transactions.
**Describe the bug**
When a delegatee was involved when charging method fee, there was a potential risk of race condition on accessing the same delegatee in parallel transactions. 

**To Reproduce**
We find this bug when code reviewing:
1. Before a transaction was inserted into transaction pool, the resource of fromAddress and toAddress would be used to judge whether there was potential conflict when executed parallel later.
2. When a delegatee was added to a delegator, the resource of the delegatee might be accessed when the delegator did not have enough tokens. However, the delegatee is not involved when judging the potential conflict.
3. So it was needed to add the ref block height when judging the potential conflict, to SDK, and avoid using the delegatee who was added lower than this ref block when charging method fee.

**Expected behavior**
N/A

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS and version: Ubuntu
 - Dotnet core version: 6.0

**Additional context**
N/A

kaiwenzhengaelf:
closes #3342, #3343


jason-hoopox (approved):

LiWang-555 (approved):
No problems found.

PR #3384, opened 2023-03-16
Fix code check failure in parallel

### LINKED ISSUE #3386: Ensure code check is successful in parallel
**Question** 
Parallel code checks will fail for the same assembly.

**Solution**
- Remove transaction pool code check college.
- Use queue processing code check to prevent concurrent.

zhxymh:
Closes #3386


src/AElf.Kernel.CodeCheck/Application/CodeCheckValidationProvider.cs
```diff
@@ -52,18 +54,37 @@ public async Task<bool> ValidateBlockAfterExecuteAsync(IBlock block)
...
+                BlockHash = blockHash,
+                BlockHeight = block.Header.Height,
+                ContractAddress = genesisContractAddress
+            }).GetSmartContractRegistrationByCodeHash.CallAsync(codeHash);
+
+            if (await _codeCheckService.PerformCodeCheckAsync(contractRegistration.Code.ToByteArray(),
+                    blockHash, block.Header.Height, contractRegistration.Category,
+                    contractRegistration.IsSystemContract, contractRegistration.IsUserContract))
+            {
+                continue;
```
gldeng, line 80: Should we call `_checkedCodeHashProvider.AddCodeHashAsync` after validation?

src/AElf.Kernel.CodeCheck/Application/CodeCheckValidationProvider.cs
```diff
@@ -12,15 +12,17 @@ internal class CodeCheckValidationProvider : IBlockValidationProvider
     private readonly ICheckedCodeHashProvider _checkedCodeHashProvider;
     private readonly IContractReaderFactory<ACS0Container.ACS0Stub> _contractReaderFactory;
     private readonly ISmartContractAddressService _smartContractAddressService;
+    private readonly ICodeCheckService _codeCheckService;
 
     public CodeCheckValidationProvider(ISmartContractAddressService smartContractAddressService,
```
gldeng, line 17: **A question (for thought only, may not be resolved in this PR):**

What's the relationship between `Provider` and `Service`? Does it make sense to have some `Provider`s calling some `Service`s while some other `Service` are calling some `Provider`s?

src/AElf.Kernel.CodeCheck/Application/CodeCheckValidationProvider.cs
gldeng: **Related to this, but not resolvable this time round:**

1. Should we call `_checkedCodeHashProvider.AddCodeHashAsync` inside `CodeCheckService.PerformCodeCheckAsync`?
2. Does it make sense to have `_checkedCodeHashProvider.AddCodeHashAsync`  depend on the block height?
3. Does it make sense to configure `RequiredAcs` in Configuration contract? Or should we make `RequiredAcs` backward compatible?

zhxymh: > Should we call `_checkedCodeHashProvider.AddCodeHashAsync` after validation?

I think it may not be good to add it here because the original data is stored according to the height, the real height is not known here, and the current height may make the data confusing. College here has a legacy: there's no place to clear the data.

zhxymh: > **Related to this, but not resolvable this time round:**
> 
> 1. Should we call `_checkedCodeHashProvider.AddCodeHashAsync` inside `CodeCheckService.PerformCodeCheckAsync`?
> 2. Does it make sense to have `_checkedCodeHashProvider.AddCodeHashAsync`  depend on the block height?
> 3. Does it make sense to configure `RequiredAcs` in Configuration contract? Or should we make `RequiredAcs` backward compatible?

1、It is not necessary to store code everywhere you do `CodeCheckService.PerformCodeCheckAsync`, like tx pool pre-validation. So `_checkedCodeHashProvider.AddCodeHashAsync` inside `CodeCheckService.PerformCodeCheckAsync` maybe not a good idea.

src/AElf.Kernel.CodeCheck/Application/CodeCheckValidationProvider.cs
zhxymh: I'm not sure about that, but I think `Provider` can only be called from its own module.

gldeng (approved):

Issue #3381, opened 2023-03-10
Make  sure CodeCheck covers all CodePatch cases


gldeng:
**Current State**
When deploying a smart contract, a user sends a proposal to for the deployment and the code will be deployed after 2/3 of the BPs have approved it. BP will manually check the code to make sure it conforms to the required security and operation standards.

**New**
In 1.3.0, the BP node client will perform CodeCheck on deployed `UserContract` automatically. BPs will automatically approve the contract if the CodeCheck is passed and  once 2/3 of the BPs approved the deployment it will be released and users can start to interact with the smart contract. One of the requirements is the deployed contract have to be correctly patched so that it can be executed securely on the blockchain.

As the CodeCheck is automatically done, we need to make sure CodeCheck covers all the cases required in CodePatch.

**Task Breakdown**
- [ ] Prepare test cases for missing coverage
- [ ] Add missed CodeCheck cases
- [ ] Refactor existing test cases to have more fine-grained test cases **(Optional)**
- [ ] Refactor CodeCheck and CodePatcher for better flexibility on usage **(Optional)**

PR #3385, opened 2023-03-17
3381 exhaustive codecheck refactor

gldeng:
Completing #3381: Refactoring CodePatcher/CodeCheck code and add unit tests to support the next phase of code changes.

**Changes:**
1. Add missed branching OpCodes
2. Disallow String.Concat
3. Disallow static fields
4. Add check for single module and single contract


src/AElf.CSharp.CodeOps/Patchers/Module/CallAndBranchCounts/Patcher.cs
```diff
@@ -0,0 +1,94 @@
...
+    public void DoPatch()
+    {
+        if (!_method.HasBody) return;
+        if (_alreadyPatched) return;
+        var processor = _method.Body.GetILProcessor();
+        processor.Body.SimplifyMacros();
+        InsertCallCountAtBeginningOfMethodBody(processor);
+        InsertBranchCountForAllBranches(processor);
+        processor.Body.OptimizeMacros();
+        _alreadyPatched = true;
```
zhxymh, line 69: Each method will create a new `MethodPatcher`, what does this state `_alreadyPatched` do?

src/AElf.CSharp.CodeOps/Patchers/Module/ResetFieldsMethodInjector.cs
```diff
@@ -1,42 +1,49 @@
...
+
     public void Patch(ModuleDefinition module)
     {
+        var checker = new ExecutionObserverProxyChecker(module);
+        var disallowedFields = module.GetAllTypes().SelectMany(t => t.Fields).Where(f => f.IsDisallowedStaticField())
+            .Where(f => !checker.IsObserverFieldThatRequiresResetting(f)).ToList();
+        if (!disallowedFields.IsNullOrEmpty())
+        {
+            throw new InvalidCodeException("Static field is not allowed in user code.");
+        }
```
zhxymh, line 24: Should there be CodeCheck logic in the CodePatch phase?

gldeng: Opened https://github.com/AElfProject/AElf/issues/3389 to track it properly.

src/AElf.CSharp.CodeOps/Patchers/Module/CallAndBranchCounts/Patcher.cs
gldeng: The intention was to prevent caller from accidentally calling the method multiple times and screw up the patching.
I think it's just a matter of how we design the contract between the caller and the callee. It's true that currently the caller follows the rule (not calling more than once) implicitly. But there's no harm that we check on the callee side explicitly.

AElf.All.sln
```diff
@@ -365,6 +365,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AElf.Kernel.FeatureManager"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AElf.Kernel.FeatureManager.Tests", "test\AElf.Kernel.FeatureManager.Tests\AElf.Kernel.FeatureManager.Tests.csproj", "{1B44277E-74EB-49B2-B8FD-05C29EE51985}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AElf.CSharp.CodeOps.UnitTests", "test\AElf.CSharp.CodeOps.UnitTests\AElf.CSharp.CodeOps.UnitTests.csproj", "{D1A00CD6-958E-4E9F-8325-354309E3029E}"
```
jason-hoopox, line 368: Unit test project end with Tests is better.

gldeng: There is an existing one `AElf.CSharp.CodeOps.Tests`. This one is the real unit tests.

zhxymh (approved):

jason-hoopox (approved):

PR #3438, opened 2023-07-14
[WIP]feature: docker start chain locally (deployment testing for developers)

aelf-lu:
- add on docker script
- modify appsettings and get methods
This docker image is used to start chain locally, it's just used to test deploy for developers.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,20 @@
+version: '3'
+services:
+  redis:
+    image: redis
+    ports:
+      - 6379:6379
+
+  aelf-node:
+    depends_on:
+      - redis
+    working_dir: /app
+    image: aelf/node
+#    command: dotnet /app/AElf.Launcher.dll --environment Docker --config.path /app
+    command: ./docker-start-chain.sh
+    ports:
+      - 6801:6800
+      - 8001:8000
+    volumes:
+      - ~/.local/share/aelf/:/root/.local/share/aelf/
+
--- a/docker-start-chain.sh
+++ b/docker-start-chain.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+password=admin123
+
+(echo "y"
+sleep 1
+echo $password
+sleep 1
+echo $password)|aelf-command create
+
+cd ~/.local/share/aelf/keys
+
+export AELF_ADDRESS=`ls -lt |awk '{if ($9) printf("%s\n",$9)}'|head -n 1|cut -d . -f1`
+export AELF_PASSWORD=password
+
+echo $AELF_ADDRESS
+echo $AELF_PASSWORD
+
+cd /app
+
+dotnet /app/AElf.Launcher.dll --environment Docker --config.path /app
--- a/src/AElf.Launcher/Dockerfile
+++ b/src/AElf.Launcher/Dockerfile
@@ -1,4 +1,17 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+#FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+FROM mcr.microsoft.com/dotnet/sdk:6.0
+
+# 安装 Node.js
+RUN apt-get update && \
+    apt-get install -y curl && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
+    apt-get install -y nodejs
+
+# 刷新软件包缓存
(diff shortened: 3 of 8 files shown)


docker-start-chain.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+password=admin123
+
+(echo "y"
+sleep 1
+echo $password
+sleep 1
+echo $password)|aelf-command create
```
gldeng, line 9: Use a fixed private key for testing. No need to generate every time.

src/AElf.Launcher/Dockerfile
```diff
@@ -1,7 +1,20 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+#FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+FROM mcr.microsoft.com/dotnet/sdk:6.0
+
+# 安装 Node.js
```
gldeng, line 4: English please.

src/AElf.Launcher/Dockerfile
```diff
@@ -1,7 +1,20 @@
...
+RUN apt-get update && \
+    apt-get install -y curl && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
+    apt-get install -y nodejs
+
+# 刷新软件包缓存
+RUN apt-get update
+
+# 安装 aelf-command 包
+RUN npm install aelf-command -g
```
gldeng, line 14: Why do we need it in the docker?

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+  redis:
+    image: redis
+    ports:
+      - 6379:6379
+
+  aelf-node:
+    depends_on:
+      - redis
+    working_dir: /app
+    image: aelf/node
```
gldeng, line 12: Don't call it this name. Maybe use `aelf/standalone-testing-node`

gldeng (approved):

src/AElf.Launcher/appsettings.Docker.json
```diff
@@ -0,0 +1,9 @@
+{
+  "ConnectionStrings": {
+    "BlockchainDb": "redis://redis:6379?db=1",
+    "StateDb": "redis://redis:6379?db=1"
```
gldeng, line 4: Can we try if "InMemory" works? See this -> https://github.com/AElfProject/AElf/blob/a5cc8ddd7509d4bc6d0a6f71cc4b253e186d85dc/src/AElf.Blockchains.MainChain/MainChainAElfModule.cs#L35-L39

src/AElf.OS/OSAElfModule.cs
```diff
@@ -25,7 +26,20 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
+            Configure<AccountOptions>(configuration.GetSection("Account"));
+        }
+        else 
+        { 
+            Configure<AccountOptions>(option => 
+            { 
+                option.NodeAccount = address; 
+                option.NodeAccountPassword = password;
+            });
+        }
```
gldeng, line 42: Why do we need this customization?

docker-start-chain.sh
aelf-lu: If we want to use a fixed private key. There are probably two ways we can do this. One is developers to install the aelf-command, and create account locally. Another is we put a fixed private key to our docker image.

src/AElf.Launcher/Dockerfile
aelf-lu: Modified.

src/AElf.Launcher/Dockerfile
aelf-lu: Now we install aelf-command and create account in the docker container. We can use a fixed private key, then we don't need to install aelf-command.

docker-compose.yml
aelf-lu: Modified.

src/AElf.OS/OSAElfModule.cs
aelf-lu: For the same reason above. Now we install aelf-command and create account in the docker container, so every time we get a new private key, we need to set it as an environment variable so that load it when the chain starts.

aelf-lu:
Migrated relevant files to aelf-developer-tools. Close this PR, and open another PR.

Issue #3549, opened 2024-05-17
The key of contract states and block executed data should be a hash value


eanzhao:
Currently, the keys are possibly too long to set to the ssdb.
Needs to shorter them.

PR #3550, opened 2024-05-17
Hashing the state key before it writing to the database

eanzhao:
- [x] Hash the state keys before writing to the database
- [x] Deal with the getting logic

Issue: #3549


eanzhao:
Seeking for another solution.

PR #3621, opened 2024-11-19
Transaction status result optimization

JimAelf:
set a the Error message to the output to infer that the transaction will never succeed. For issue #3623


eanzhao (approved):

chopin-fan (approved):
have checked

PR #3641, opened 2026-04-14
Feature: Add block pruning module for historical blockchain data cleanup

jason-aelf:


eanzhao:
There is also a protocol-level compatibility issue when the requester's `PreviousBlockHash` has already been pruned on the serving node.

Once the header for `firstHash` is gone, `GetBlockHashesAsync()` returns an empty list immediately. On its own that is fine, but on the download side `DownloadBlockCount == 0` is treated as "job finished" and the sync job is removed. In other words, a lagging node talking only to pruning peers can silently stop the download attempt instead of learning that this peer no longer has the requested historical range.

That makes the interaction look like a successful empty response rather than "history unavailable". I think the sync protocol needs a distinguishable outcome here, or a fallback path, otherwise long-offline nodes can get stuck depending on which peers they hit.

eanzhao:
I think there is a correctness issue in the historical sync path after pruning.

`BlockPruningService` intentionally keeps the height-to-hash index but removes the corresponding historical `BlockHeader` / `ChainBlockLink` entries for pruned heights. The problem is that `GetBlocksWithTransactionsAsync()` still goes through `GetBlocksInBestChainBranchAsync()` -> `GetBlockHashesAsync()`, and that traversal still assumes those historical links exist.

In particular, if a fresh peer asks for blocks starting from genesis after pruning has already removed height 2/3/... data, `GetBlockHashesAsync()` can resolve `first.Height`, compute a target height, and then fail when it tries to walk back through a pruned `ChainBlockLink` before the later `TakeWhile(block != null)` safeguard in `BlockchainServiceExtensions` is reached. That turns a normal block request into an exception in the gRPC request path instead of a clean partial response.

So a pruning node can become unable to serve historical sync from genesis/new peers. I think this path needs an explicit "history pruned" behavior, or a traversal that stops cleanly once historical links are missing, not just null filtering at the very end.

